PR #8, opened 2022-04-22
MAINT: CI updates

misialq:
Similarly to https://github.com/bokulich-lab/q2-assembly/pull/19, this PR adds a coverage testing step to the CI.


misialq:
CI appears broken as there are not tests available yet - will be fixed in the next PR.

PR #12, opened 2022-06-11
ci: clean up coverage tasks

thermokarst:


thermokarst:
x-ref https://github.com/bokulich-lab/q2-assembly/pull/20 - this plugin will have the same issue. we can sort it out in this pr, or another, your call @misialq

misialq:
Hey @thermokarst, many thanks for these updates - these look great! I think I'd merge this one and take care of the other issue in a separate PR (although I am not entirely sure what that would entail, as pointed out here: https://github.com/bokulich-lab/q2-assembly/issues/16#issuecomment-1158705430). I'll also update the CI of the other plugins to follow the changes introduced here.

thermokarst:
A separate PR sounds good. RE your question to Evan in that q2-assembly post - the answer is "yes" - this is a problem in any of your plugins - you'll need to use conda for _all_ dependencies in order to correctly build and publish on Library.

I think I'm set on this end for this PR, so feel free to merge if/when you're ready. Thanks!

PR #13, opened 2022-06-22
dep: install checkm and pplacer via conda

thermokarst:


thermokarst:
@misialq - this PR is a start for getting a pure-conda recipe in order. Unfortunately you'll have to sort out getting a functional pplacer build for macos. It looks like there has been a bit of churn on this topic over at bioconda/bioconda-recipes.

ebolyen:
Yeah it also looks like very few have even attempted a source build, which is actually pretty odd for bioconda. I'm going to see if I can get such a recipe to work, and then a lot of that mess ought to be solved.

misialq:
Thanks for looking into that @thermokarst and @ebolyen! 🙇

ebolyen:
So some bad news:

I have half of a recipe that _might_ work, except for a downstream dependency issue in pplacer itself.
pplacer has a dependency on `sqlite3=1.6.3` (ocaml). OPAM used to know this as `sqlite3-ocaml` and around 2014 was probably the last time this version was publicly accessible. I could look into how opam repositories are handled, and then create a git clone of an ancient tag to create an artificial opam repository. This is already done for some other magical dependencies: https://github.com/matsen/pplacer-opam-repository/tree/master/packages

But I think the worse headline news is, this package cannot actually be compiled except by a handful of developers who happen to have this ancient version in their opam cache. Worse, we shouldn't expect any meaningful updates as [pplacer is no longer developed or maintained](https://github.com/matsen/pplacer/issues/366#issuecomment-461007465).

I might be able to work around these things, but we'd be mocking out ancient versions of things to resuscitate a package that is already quite dead.
image


---

Does it make sense to go through that? Given the dependency on pplacer through CheckM, it seems like there may be a need for some kind of `vsearch`-style reimplementation of pplacer or maybe just a fork? Alternatively, it looks like there may be some alternatives: http://genesis-lib.org/ (although I can't tell at first glance if it's the same kind of problem, but it does mention RAxML-EPA as an alternative).

ebolyen:
It looks like [epa-ng](https://github.com/Pbdas/epa-ng) might be a viable alternative, although that just pushes the issue up to CheckM.

misialq:
Oh my, what a, pardon me, mess... 😭  Thanks for going through that @ebolyen! I discussed today with @nbokulich, and we both feel like this is way too much work for it to be worth it... We've had enough trouble with CheckM already (like python 3.8 compatibility issues, or the outdated DB) - let's not invest more effort into that. We are thinking that instead of doing that, we could put CheckM into its own, separate plugin which would not become part of the Library - that way we can keep those dependencies separate, install pplacer the way it is now being installed and be done with it. Instead we could invest time into integrating some more modern QC evaluation tools into moshpit (e.g. we have discussed [BUSCO](https://busco.ezlab.org/)). What do you think?

PR #31, opened 2023-04-06
FIX: allow Kraken2 confidence to be 1.0

### LINKED ISSUE #30: `classify-kraken` parameter `confidence` should allow confidence of 1.0
I'm experimenting with difference confidence values for this classifier, and my understanding is that 1.0 is a valid value here but is disallowed by the plugin. I think the type needs to be updated to `Float % Range(0, 1, inclusive_end=True)`. 

```
 Parameter 'confidence' received 1 as an argument, which is incompatible with parameter type: Float % Range(0, 1)
```

misialq:
Closes #30.

### DIFF
--- a/q2_moshpit/plugin_setup.py
+++ b/q2_moshpit/plugin_setup.py
@@ -102,7 +102,7 @@ plugin.methods.register_function(
     },
     parameters={
         'threads': Int % Range(1, None),
-        'confidence': Float % Range(0, 1),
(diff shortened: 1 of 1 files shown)


misialq:
Here you go @gregcaporaso! Unless you have any comments here, I'll merge that.

gregcaporaso (approved):
Looks good, thank you!

PR #62, opened 2023-09-06
ENH: Parallelize classify kraken2

Oddant1:
Work on parallelizing `classify_kraken2` in the same manner as `assembly_megahit` in bokulich-lab/q2-assembly#46


q2_moshpit/kraken2/classification.py
```diff
@@ -47,15 +51,92 @@ def _construct_output_paths(
...
+def classify_kraken2_helper(
         seqs, common_args
 ) -> (Kraken2ReportDirectoryFormat, Kraken2OutputDirectoryFormat):
     if isinstance(seqs, MAGSequencesDirFmt):
         manifest = None
     else:
         manifest: Optional[pd.DataFrame] = seqs.manifest.view(pd.DataFrame)
 
-    base_cmd = ["kraken2", *common_args]
+    base_cmd = ["kraken2", *common_args, '--memory-mapping']
```
Oddant1, line 139: This `--memory-mapping` flag prevents the database from being loaded into RAM for every thread. It slows things down compared to loading the database for every thread, but it prevents us from needing to load a ~60GB database into RAM a million times. For now we'll have it here to be safe, but we can definitely be smarter about when we do and do not need it and how we handle the database (/dev/shm for instance).

Oddant1:
Blocked by qiime2/q2-demux#145 and qiime2/q2cli#306

Oddant1:
Blocked by qiime2/qiime2#723

misialq:
Hey @Oddant1, I just invited you to our moshpit GitHub team - if you accept the invite the workflows should self-approve from now on :)

Oddant1:
blocked by bokulich-lab/q2-assembly#46

q2_moshpit/kraken2/classification.py
gregcaporaso: I see the action's current default is False (no memory mapping) - do you think it should stay that way with the addition of parallel support? I could see arguments either way.

colinvwood:
Passing locally on both a Mac and linux.

colinvwood (approved):

colinvwood:
@misialq we're ready for a merge on this, CI is not syncing as previously discussed

gregcaporaso:
Thanks for all of the work on this @Oddant1, and for doing the review @colinvwood and @cherman2!

PR #66, opened 2023-09-27
TST: adds more tests to kraken2_to_features

ebolyen:
Sorry it took a while to come up with a sane strategy for this.

This all seems to work alright, and I can confirm the tests where failing in useful/obvious ways as I was developing it, so they do in fact do what they say.

I did also find an edge case, where a report that ends with an infraclade rank would be skipped, rather than walking up to the last valid rank:
```
         R;root
         D;  a
        D1;    a.s
         K;    a.b
        K1;      a.b.s
         P;        a.b.s.a
         D;  b
         K;    b.a
         P;      b.a.a
        P1;        b.a.a.a
```
That's been corrected, and the coverage report looks pretty good locally. The only few issues with branch coverage are some protective branches which never jump in (I believe) an impossible way.


ebolyen:
This seems to be a victim of the stale CI, will be working on that next.

ebolyen:
So in order to get the CI working, we do need a passing test.
I'm merging this to get that going, but a review would still be potentially warranted.

q2_moshpit/kraken2/select.py
```diff
@@ -162,9 +165,16 @@ def _kraken_to_ncbi_tree(df):
...
+    # It is possible for the last row to be an infra-clade tip,
+    # so walk backwards up the stack until a standard node (with length)
+    # is found
+    while stack and parent_node.length == 0:
+        _, parent_node = stack.pop()
+
     if parent_node.children:
         parent_node.children[0].is_actual_tip = True
 
+    print(tree)
```
ebolyen, line 177: Well. I'll need to fix that. Expect another PR from me @misialq, but hopefully with a working CI.

Issue #63, opened 2023-09-07
add `SampleData[Contigs]` as input type to `classify-kraken2`


gregcaporaso:
Output should `KrakenReport % Property('contigs')` and `KrakenOutput % Property('contigs')`

colinvwood:
closed by #69

PR #69, opened 2023-09-29
Classify contigs

colinvwood:
allow classify_kraken2 to take contigs as input


colinvwood:
blocked by bokulich-lab/q2-types-genomics/#54 and bokulich-lab/q2-types-genomics/#55

q2_moshpit/kraken2/select.py
```diff
@@ -137,7 +137,7 @@ def _kraken_to_ncbi_tree(df):
     for _, row in df.iterrows():
         r = row['rank']
         label = row['name']
-        otu = str(row['ncbi_tax_id'])
+        otu = str(row['taxon_id'])
```
ebolyen, line 140: :+1:

q2_moshpit/kraken2/tests/test_selection.py
```diff
@@ -236,7 +236,7 @@ def make_dirfmt(self, string, coverage=False):
                      pad1=0,
                      pad2=0,
                      rank=rank,
-                     ncbi_tax_id=tax_id,
+                     taxon=tax_id,
```
ebolyen, line 239:
```suggestion
                     taxon_id=tax_id,
```
to match, right?

colinvwood: yes fixed

colinvwood:
closes #63

q2_moshpit/kraken2/classification.py
```diff
@@ -50,7 +51,7 @@ def _construct_output_paths(
 def _classify_kraken2(
         seqs, common_args
 ) -> (Kraken2ReportDirectoryFormat, Kraken2OutputDirectoryFormat):
-    if isinstance(seqs, MAGSequencesDirFmt):
+    if type(seqs) is MAGSequencesDirFmt or type(seqs) is ContigSequencesDirFmt:
```
misialq, line 54: You could also do `isinstance(seqs, (MAGSequencesDirFmt, ContigSequencesDirFmt))` for simplicity (or was the intention to do a strict type checking?)

q2_moshpit/kraken2/tests/test_classification.py
```diff
@@ -321,5 +321,121 @@ def test_classify_kraken_action(self, p1):
         p1.assert_called_with(ANY, exp_args)
 
 
+class TestKraken2ClassifyContigs(unittest.TestCase):
```
misialq, line 324: To keep things consistent here, maybe you could either split the tests above into respective classes for reads and  MAGs or just include the tests below into the class that already contains the other ones? The way it is now makes the contig classification seem as if it was somehow special compared to the other ones...

colinvwood: Hey @misialq, I was going to add tests for reads and mags that prove functionality as I have for the contigs, but there wasn't enough time for this release. I can go ahead and spilt each sequence type into its own test class, you're right that that'll be more consistent.

q2_moshpit/kraken2/classification.py
colinvwood: I didn't know you could use `isinstance` that way, thanks!

colinvwood:
can confirm that all tests are passing locally

q2_moshpit/kraken2/tests/test_classification.py
```diff
@@ -271,54 +340,121 @@ def test_classify_kraken_reads(self, p1, p2, p3, p4, p5):
...
+        }
+
+        output_views = outputs.reports.iter_views(pd.DataFrame)
+        for path, df in output_views:
+            sample_id = str(path).rsplit('.output.txt')[0]
+
+            # no sequences are unclassified
+            self.assertNotIn('U', list(df['classification']))
+
+            # all classifications are correct
```
gregcaporaso, line 441: You could simplify this with something like:
```
self.assertEqual(pd.unique(df['taxon_id']), [sample_id_to_ncbi_id[sample_id]])
```

gregcaporaso (changes requested):
To be honest, the `unittest.mock` stuff kind of blows my mind, but it all looks reasonable to me. The test of the public-facing function looks good - I have one comment about a missing check in there that should be addressed, and the other comments are just about creating issues and a suggestion about how one of the tests is written. 

Nest steps post-alpha (let's note these in the issue that you create): 
1. We need to come back to these tests after the alpha release to fill in the missing ones now that you've added the test data/database (thanks for doing that!), and we should probably test edge cases at that time, like empty inputs for some samples and a sample with no hits to the database. We should also test merging some of your test samples to confirm that multiple expected taxon ids show up in that case. 
2. We need to make sure that we have replaced all parsing of sample ids in here.

ebolyen:
Just an FYI that this needs to be merged before we can release, as it is paired with https://github.com/bokulich-lab/q2-types-genomics/pull/55

ebolyen:
Merging because the `KeyError` won't be a problem, we just haven't gotten a build with the updates in types-genomics yet.

PR #71, opened 2023-10-03
BUG: fixing test failure with sorting order for MAGs

lizgehret:
Fix for [this intermittent test failure](https://github.com/qiime2/distributions/actions/runs/6386509986/job/17338770423?pr=27#step:8:341) b/c glob doesn't maintain a consistent sorting order.


ebolyen:
Tested locally and this works.

The issue was that on certain filesystems (such as the mysterious ones used by GH Action runners) `glob.glob` will return things in an arbitrary order. By sorting the glob first, it's easier to assert things in order.

The functionality was correct either way, but this prevents test failures on differing filesystem traversals.

ebolyen (approved):

ebolyen:
Also lint seems to be fine :)

PR #72, opened 2023-10-04
CI: updates CI to new system

ebolyen:
Same as https://github.com/bokulich-lab/q2-types-genomics/pull/53


ebolyen:
Well.... coverage is working I guess. @misialq, is this ok to merge? This PR happens to have less lines, which changed the number covered thereby breaking the threshold. I don't know if it's a fixed percent (98%) that is the target, or if it just complains when the number goes down.

ebolyen:
Well, thinking about it more. The code that is or isn't failing codecov is unrelated to this PR, which can now measure it. And CI was already failing. So I'll merge this now and we can fix up coverage if needed.

But it seems like codecov just hasn't had a good report in a while.

misialq:
Yeah, it's been a while since we've seen acceptable report here so let's just observe from now on - thanks! 🙏

PR #79, opened 2023-10-09
change how contig and mag files are collected in _classify_kraken2

### LINKED ISSUE #76: BUG: os.glob should also search for .fa files
we are only finding `.fasta` files [here](https://github.com/bokulich-lab/q2-moshpit/blob/2b1196620e604f6775160af276d7cc3b41bf8f3e/q2_moshpit/kraken2/classification.py#L84) when we should also be getting `.fa` files.

This probably hasn't been noticed because metabat2 (probably) outputs its mags in .fasta formats and contigs were only recently added as inputs.

Update: megahit outputs .fa files so none of the contigs generated with that tool are being discovered.

colinvwood:
Fixes #76

### DIFF
--- a/q2_moshpit/kraken2/classification.py
+++ b/q2_moshpit/kraken2/classification.py
@@ -6,7 +6,6 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 import os
-import pathlib
 import subprocess
 from copy import deepcopy
 from typing import Union, Optional
@@ -82,11 +81,9 @@ def _classify_kraken2(
             path_function = get_paths_for_reads
         else:
             # we got contigs or MAGs
-            def view_to_paths(view):
-                path = view[0]
-                stem = path.stem
-                absolute = str(pathlib.Path(seqs.path) / path)
-                return stem, absolute
+            def view_to_paths(arg):
+                relpath, _ = arg
+                return relpath.stem, str(seqs.path / relpath)
 
             iterate_over = map(
                 view_to_paths, seqs.sequences.iter_views(DNAFASTAFormat)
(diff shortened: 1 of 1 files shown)


q2_moshpit/kraken2/classification.py
```diff
@@ -75,14 +76,22 @@ def get_paths_for_contigs(contig_id, fp):
...
         else:
-            iterate_over = (
-                (os.path.basename(fp).split(".")[0], fp)
-                for fp in sorted(glob.glob(os.path.join(seqs.path, "*.fasta")))
+            # we got contigs or MAGs
+            def view_to_paths(view):
+                path = view[0]
+                stem = path.stem
+                absolute = str(pathlib.Path(seqs.path) / path)
+                return stem, absolute
```
ebolyen, line 89: I don't think the variables are adding a whole lot (maybe `absolute` is, but I dunno if it's worth the line):
```suggestion
                return relpath.stem, str(seqs.path / relpath)
```

q2_moshpit/kraken2/classification.py
```diff
@@ -75,14 +76,22 @@ def get_paths_for_contigs(contig_id, fp):
...
+            # we got reads - use the manifest
             iterate_over = manifest.iterrows()
             path_function = get_paths_for_reads
         else:
-            iterate_over = (
-                (os.path.basename(fp).split(".")[0], fp)
-                for fp in sorted(glob.glob(os.path.join(seqs.path, "*.fasta")))
+            # we got contigs or MAGs
+            def view_to_paths(view):
+                path = view[0]
```
ebolyen, line 86: This adds a tiny amount of checking, and makes it a little more clear that the input is a tuple, rather than a view which has some kind of get-item maybe?
```suggestion
            def view_to_paths(arg):
                relpath, _ = arg
```

ebolyen (approved):

PR #80, opened 2023-10-10
ENH: Busco main plot. Bars with constant height.

### LINKED ISSUE #78: ENH: On busco's main plot, make the facet plots vary in height depending on the number of mags in each sample.
**Problem:**
In the main visualization, all samples are allocated the same amount of space for plotting independently of the number of mags they contain. The height of the bars representing each of the MAGs in the plot is then adjusted to fit this predefined space. This leads to bars having different heights since different samples have different numbers of MAGs. This has been found to be aesthetically unpleasing and perhaps even cumbersome to the interpretation of the plot. 

**Solution:**
Make it so all bars have the same height thereby adjusting the height of the per-sample bar plots to account for different numbers of MAGs/bars.

**Possible Implementations**
*Plan A:* If it's possible leave the implementation as is (using the facet feature of the plotting library).
*Plan B:* We change the way the plot is constructed to allow bars to have the same size.

Sann5:
**Closes #78**

- Controls the height of the plot by assigning a constant height to each bar.
- Controls the vertical spacing between subgraphs.
- Adjusts tests and calls to this function accordingly.

### DIFF
--- a/q2_moshpit/busco/tests/data/plot_as_dict.json
+++ b/q2_moshpit/busco/tests/data/plot_as_dict.json
@@ -473,6 +473,7 @@
           "y": "independent"
         }
       },
+      "spacing": 20,
       "spec": {
         "encoding": {
           "color": {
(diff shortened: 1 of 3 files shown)


misialq (approved):
Exactly what I had in mind - thanks! 🚀

PR #86, opened 2023-10-17
make bin-contigs collect .fasta contigs too, other small changes

### LINKED ISSUE #85: BUG: bin-contigs-metabat only collects .fa files
Similar to #76, at several places in `q2_moshpit/metabat2/metabat2.py` only `.fa` files are collected when `.fasta` files should also be collected.

colinvwood:
Fixes #85


misialq (approved):
Great, thanks!

PR #89, opened 2023-10-18
Add functionality tests for classify-kraken2

### LINKED ISSUE #73: add tests that prove functionality for classify-kraken2 with reads as input and with MAGs as input
We need to come back to these tests after the alpha release to fill in the missing ones (commented as TODO in `kraken2/tests/test_classification`) using the same database that's used by the contig tests. We should probably also test edge cases at that time, like empty inputs for some samples and a sample with no hits to the database. We should also test merging sequences from multiple test samples to confirm that multiple expected taxon ids show up in the classification results.

see discussion in #69

colinvwood:
Closes #73


misialq:
Hey @colinvwood, not sure whether you are done with this one, but it seems like the `test_reads` test does not assert anything (the filtered `output_views` is empty) - you can see it in the coverage report. Could you please check? Thanks! 🙏

misialq (approved):
Great, thanks! 🚀

PR #93, opened 2023-11-02
ENH: Added prodigal to moshpit

### LINKED ISSUE #5: Implement gene prediction from dereplicated MAGs
We need an action (`predict-genes-prodigal`) supporting gene prediction on dereplicated MAGs.

**Acceptance criteria:**

- uses metaProdigal ([link](https://github.com/hyattpd/Prodigal))
- uses `FeatureData[MAG]` as input
- outputs `GenomeData[Loci]` +  `GenomeData[Genes]` + `GenomeData[Proteins]` + `FeatureTable[Frequency]`

Sann5:
**Closes #5**

Includes test suite for the wrapper implementation.


q2_moshpit/plugin_setup.py
```diff
@@ -560,3 +562,48 @@
...
+plugin.methods.register_function(
+    function=q2_moshpit.prodigal.predict_genes_prodigal,
+    inputs={
+        'mags': FeatureData[MAG]
+    },
+    input_descriptions={
+        'mags': 'MAGs for which one wishes to predict genes.'
+    },
+    parameters={
+        "translation_table_number": Int % Range(1, 26)
```
VinzentRisch, line 575:
```suggestion
        "translation_table_number": Str % Choices(['1', '2', '3', '4', '5', '6', '9',
                                                   '10', '11', '12', '13', '14', '15',
                                                   '16', '21', '22', '23', '24', '25'])
```
It really bugs me that you can input a parameter that leads prodigal to fail.
With this change, only the working parameters can be inputted. If a wrong one is given, there is directly an error message, and you don't have to open the log to see what the problem is.
Unfortunately, it's not possible to use `Choices` with integers, but it does not change anything for the user if the parameters are strings.

VinzentRisch (changes requested):

q2_moshpit/citations.bib
```diff
@@ -70,3 +70,23 @@ @article{manni_busco_2021
 	pages = {4647--4654},
 	file = {Manni et al. - 2021 - BUSCO Update Novel and Streamlined Workflows alon.pdf:/Users/santiago/Zotero/storage/SQ2VFGPF/Manni et al. - 2021 - BUSCO Update Novel and Streamlined Workflows alon.pdf:application/pdf},
 }
+
+@article{hyatt_prodigal_2010,
```
misialq, line 74: Nice, thanks for remembering about that one!

q2_moshpit/citations.bib
```diff
@@ -70,3 +70,23 @@ @article{manni_busco_2021
...
+	abstract = {The quality of automated gene prediction in microbial organisms has improved steadily over the past decade, but there is still room for improvement. Increasing the number of correct identifications, both of genes and of the translation initiation sites for each gene, and reducing the overall number of false positives, are all desirable goals.},
+	number = {1},
+	urldate = {2023-10-23},
+	journal = {BMC Bioinformatics},
+	author = {Hyatt, Doug and Chen, Gwo-Liang and LoCascio, Philip F. and Land, Miriam L. and Larimer, Frank W. and Hauser, Loren J.},
+	month = mar,
+	year = {2010},
+	keywords = {Bias Score, Codon Position, Halobacterium Salinarum, Ribosomal Binding Site, Translation Initiation Site},
+	pages = {119},
+	file = {Full Text PDF:/Users/santiago/Zotero/storage/ST2FJXSD/Hyatt et al. - 2010 - Prodigal prokaryotic gene recognition and transla.pdf:application/pdf;Snapshot:/Users/santiago/Zotero/storage/3YPISIE8/1471-2105-11-119.html:text/html},
```
misialq, line 90: Maybe you can remove that attribute? It points to your personal files 🙈

q2_moshpit/plugin_setup.py
```diff
@@ -560,3 +562,48 @@
...
+    },
+    parameter_descriptions={
+        'translation_table_number':
+            'Translation table to be used to '
+            'translate genes into a sequence of amino '
+            'acids. See '
+            'https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi '
+            'for reference. NOTE: not all tables are supported. '
+            'Supported tables: 1-6, 9-16 and 21-25.'
```
misialq, line 584: With the change from @VinzentRisch, this line won't be necessary :)

q2_moshpit/plugin_setup.py
```diff
@@ -560,3 +562,48 @@
...
+        ('loci', GenomeData[Loci]),
+        ('genes', GenomeData[Genes]),
+        ('proteins', GenomeData[Proteins])
+    ],
+    output_descriptions={
+        'loci': "Gene coordinates files (one per MAG) listing the location of "
+                "each predicted gene as well as some additional scoring "
+                "information. File format: 'gff'. See "
+                "https://github.com/The-Sequence-Ontology/Specifications/blob/"
+                "master/gff3.md for reference.",
```
misialq, line 596: No need to provide the format info - this is implied by the semantic type.

q2_moshpit/prodigal/prodigal.py
```diff
@@ -0,0 +1,73 @@
...
+    # Pattern of fasta files in input dir
+    pattern = re.compile(
+        r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-4[0-9a-fA-F]{3}-"
+        r"[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}\.(fa|fasta)$"
+    )
+
+    # Look for fasta files in input dir
+    fasta_files = [
+        file for file in os.listdir(mags.path) if pattern.match(file)
+    ]
```
misialq, line 37: There is no need for the pattern matching here: the semantic type is taking care of that so when you receive the `MAGSequencesDirFmt` it is guaranteed it will only have the right FASTA files so you can directly use them.

q2_moshpit/prodigal/prodigal.py
```diff
@@ -0,0 +1,73 @@
...
+        d = os.path.join(genes.path, f"{file_id}_genes.fasta")
+
+        # Adjust command and run
+        cmd = cp.deepcopy(base_cmd)
+        cmd.extend([
+            "-i", i,
+            "-o", o,
+            "-a", a,
+            "-d", d
+        ])
```
misialq, line 69: This can be simplified by directly passing all the paths (`os.path...`) in the `extend` method: since you are not using the `i, o, a, d` variables anywhere else, we may as well get rid of them to streamline the code.

misialq (changes requested):
Looking good, @Sann5! See some comments below :) I still need to take it for a spin though :D

q2_moshpit/prodigal/prodigal.py
Sann5: I see. then I'll just look for files that end with fa or fasta to collect the paths. There is an exception clause right after these lines that goes off when no files are found. Should I remove that too seeing that this is something that qiime should also control for?

misialq: Actually, that's a good point - that should not be needed, you're right 👌.

misialq (approved):
Works like a charm - thanks @Sann5! 🚀

PR #94, opened 2023-11-09
ENH: action to fetch eggnog databases

### LINKED ISSUE #75: ENH: add action to fetch eggnog/diamond databases
The `eggnog`-specific actions require reference databases as inputs which at the moment need to be manually created/fetched by the user. It is not immediately clear how those should be constructed and/or what actually should be included in either of those.

Let's add an action `fetch-eggnog-db` which would grab the latest version of the entire eggnog database (using the `download_eggnog_data.py` tool provided by eggnog itself) and create the two ReferenceDB artifacts required by those actions.

**Additional context**
We could later add a `build-eggnog-db` action which could be used to construct custom eggnog databases. This has lower priority, though, as there is a comprehensive DB already available through download.

Sann5:
Closes part of #75. Issue 75 specifies that two actions should be created. This PR contains one of those two actions—namely the `fetch_eggnog_db` action. There will be a separate PR for the action to fetch the Diamond DB. 

In a nutshell, the action downloads several files and creates an artifact with them. If you wish to try it locally run the following.

First, clone the repo and checkout the PR branch:
```basg
git clone git@github.com:bokulich-lab/q2-moshpit.git
gh pr checkout 94
```

Now you can run the action locally. Be aware that this will download ~40 Gb of data and you need around 80 Gb of free disk space for it to run successfully (see comments below for more info on this).
```bash
qiime moshpit fetch-eggnog-db --o-eggnog-db eggnog_db.qza --verbose
```

Alternatively, you can run the tests for this action (nothing will be downloaded).
```bash
cd <you_local_path_to>/q2-moshpit/q2_moshpit/eggnog/tests
pytest -vv -W "ignore" -k fetch_eggnog_db
```


Sann5:
Do not merge yet! I have a weird error I have to look into when trying the action out :)

Sann5:
Ok, so I found the issue. The files are downloaded to temporary files and after the action concludes successfully qiime copies these files in the temporary directories to the actual output `qza` files. This copy operation implies that you need to double the amount of memory than the actual size of the download (since the same files exist in two different locations even if only briefly). Also, because the files are so big it takes a considerable amount of time to copy them so this adds some run time on top of the download time.

misialq (changes requested):
Looks great, thanks @Sann5! I tried it out - all worked as expected. Just some small comments below (trying to make it a bit more future-proof).

q2_moshpit/eggnog/tests/test_method.py
```diff
@@ -85,3 +85,21 @@ def test_small_good_hits(self):
...
+        # Call function. Patching will make sure nothing is
+        # actually ran
+        eggnog_db, diamond_db = fetch_eggnog_db()
+
+        # Check that command was called in the expected way
+        cmd = [
+            "download_eggnog_data.py", "-y", "--data_dir", str(eggnog_db.path)
+        ]
+        subp_run.assert_called_once_with(cmd, check=True)
+        mv.assert_called_once_with(src=ANY, dst=ANY)
```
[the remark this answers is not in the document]
Sann5: Sorry. In the past, I haven't always had access to the temporary paths created during the call, so I mistakenly assumed that I didn't have access to them here, but I do. I have updated the test to check for the call with the exact paths.

Sann5:
**PR update**
- Last few commits modify the action such that it only fetches the the eggnog DB
- Action for fetching the Diamond DB will be done in a separate PR

q2_moshpit/plugin_setup.py
```diff
@@ -379,6 +379,25 @@
...
+    function=q2_moshpit.eggnog.fetch_eggnog_db,
+    inputs={},
+    parameters={},
+    outputs=[("eggnog_db", ReferenceDB[Eggnog])],
+    output_descriptions={
+        "eggnog_db": "Artifact containing the eggNOG annotation "
+                     "database (eggnog.db, eggnog.taxa.db, "
+                     "eggnog.taxa.db.traverse.pkl)"
+    },
+    name="Fetch one of the databases necessary to run eggNOG.",
```
misialq, line 392: Maybe we can be a bit more specific and say "to run the eggnog-annotate action"? With all those databases it is becoming confusing which is needed where 😅

q2_moshpit/plugin_setup.py
```diff
@@ -379,6 +379,25 @@
...
+    description="Downloads eggnog reference database  "
+                "using the `download_eggnog_data.py` script from eggNOG. "
+                "Here, this script downloads 3 files, specifically: "
+                "eggnog.db (39Gb), "
+                "eggnog.taxa.db (265Mb), and"
+                "eggnog.taxa.db.traverse.pkl (6.3Mb)."
```
misialq, line 398: Again, let's not mention the sizes here (or anywhere else) as they may change in the future. Maybe you can say instead that the entire database requires at least 40 GB of storage space? (or however much you think it does)

misialq (changes requested):
Hey @Sann5, looks good! 🥇  Just some cosmetic changes to the descriptions - see below.

q2_moshpit/plugin_setup.py
Sann5: I need 80 since it needs to copy the files once everything is downloaded :)

misialq (approved):
Merge away, @Sann5! 🚀

PR #100, opened 2023-11-23
ENH: Fetching action for full Diamond database

### LINKED ISSUE #97: ENH: Fetching action for full Diamond database
### Is your feature request related to a problem?
No. It would be nice to have an action that downloads the Diamond database for downstream Eggnog analyses. 

### Describe the solution you'd like
This could be done by simply calling the script and piping it with the appropriate yes/no answers to indicate which database should be downloaded. For example:

```bash
printf "n\nn\ny" | download_eggnog_data.py -s --daata_dir . 
```

### Additional considerations
- Care should be taken that the output artifact can be accessed in the downstream actions.
- Care should be taken an appropriate Semantic Types is used for the generated database artifact. See this issue bokulich-lab/q2-types-genomics#64

Sann5:
## Closes #97 (ENH: Fetching action for full Diamond database)
Action to download the DIamond database and create an artifact from it. 

- [x] **NOTE**: It includes the changes from PR #94 so it makes sense to merge that PR first.

### Run it yourself!
If you wish to try it locally run the following (it is assumed that you have a working environment). 

First, clone the repo and checkout the PR branch:
```bash
git clone git@github.com:bokulich-lab/q2-moshpit.git
gh pr checkout 100
```

Now you can run the action locally. Be aware that this will download ~8.6 Gb of data and you need around 16 Gb of free disk space for it to run successfully.
```bash
qiime moshpit fetch-diamond-db --o-diamond-db diamond.qza --verbose
```

Alternatively, you can run the tests for this action (nothing will be downloaded).
```bash
cd <you_local_path_to>/q2-moshpit/q2_moshpit/eggnog/tests
pytest -vv -W "ignore" -k fetch_diamond
```


q2_moshpit/eggnog/tests/test_method.py
```diff
@@ -102,3 +104,26 @@ def test_fetch_eggnog_db(self, subp_run):
...
+        # Call function. Patching will make sure nothing is
+        # actually ran
+        diamond_db = fetch_diamond_db()
+
+        # Check that command was called in the expected way
+        cmd = [
+            'echo "Starting download..." && '
+            f'cd {str(diamond_db)} && wget -nH --user-agent=Mozilla/5.0 '
+            '--relative --no-parent --reject "index.html*" --cut-dirs=4 '
+            '-e robots=off -O ref_db.dmnd.gz '
```
misialq, line 119: You can remove most of those flags as they do not really apply here:
- `--user-agent`: impersonating Firefox should not be necessary - the download works without it
- `--relative`, `--no-parent`, `-nH`, `--reject ...` and `--cut-dirs=4` are also not so relevant as you are specifying the output file name directly

Please try without those and only keep the ones that seem to be necessary.

misialq: Ooop, I meant to make this comment in the actual code... Still, the point remains the same.

q2_moshpit/eggnog/_method.py
```diff
@@ -157,3 +158,35 @@ def fetch_eggnog_db() -> EggnogRefDirFmt:
 
     # Return objects
     return eggnog_db
+
+
+def fetch_diamond_db() -> DiamondDatabaseDirFmt:
```
misialq, line 163: For the sake of clarity, could you create a new Python file in this submodule (`dbs.py`) and move all the code relevant to fetching eggnog databases there? Maybe you can also move the exisitng `fetch-eggnog-db` action there.

q2_moshpit/_utils.py
```diff
@@ -9,7 +9,7 @@
 from typing import List
 
 
-def run_command(cmd, env=None, verbose=True, pipe=False):
+def run_command(cmd, env=None, verbose=True, pipe=False, **kwargs):
```
misialq, line 12: Uh oh, let's not do that - it introduces security concerns, as technically it would allow you to run anything in the users' system, _particularly_ if you use the _shell_ flag - see below.

q2_moshpit/eggnog/_method.py
```diff
@@ -157,3 +158,35 @@ def fetch_eggnog_db() -> EggnogRefDirFmt:
...
+    """
+
+    # Initialize output objects
+    diamond_db = DiamondDatabaseDirFmt()
+
+    # Define command.
+    # 'printf "n\nn\ny" will answer the prompts thrown by
+    # download_eggnog_data.py, effectively only downloading the
+    # Diamond DB
+    cmd = [
```
misialq, line 177: We need to break this down a little - there is no need to run that as a single command:
- move the `echo` statements to Python print statements
- break your single call to `subprocess.run` into two: one for `wget` and another one for `gunzip`
- remove the `shell=True` - it's a practice generally advised against: running commands through shell has its own security concerns (https://docs.python.org/3/library/subprocess.html#security-considerations)

misialq (changes requested):
Hey @Sann5, it works! See some comments below though :)

q2_moshpit/_utils.py
Sann5: Sorry for the dirty workaround. I need to set the working directory for the command. This could be done simply by including `cd wd &&...` but it can also be done with the `cwd` parameter in `subproccess.run`. What do you think about adding this to `run_command` function? It might be useful in future use cases. 

I also added this `colorify` function which colors the print statement to the terminal. I think it's neat since it makes it easier to read the output. Let me know what you think. I got it from the eggnog scripts so I guess if we use them we should give them credits somewhere.

Sann5:
I also added this `colorify` function to the `q2_moshpit/_utils.py` which colors the print statement to the terminal. I think it's neat since it makes it easier to read the output. Let me know what you think. I got it from the eggnog scripts so I guess if we use them we should give them credits somewhere.

q2_moshpit/eggnog/_dbs.py
```diff
@@ -0,0 +1,90 @@
...
+    `download_eggnog_data.py` script from eggNOG. Here, this
+    script downloads 3 files amounting to 47Gb in total.
+    """
+
+    # Initialize output objects
+    eggnog_db = EggnogRefDirFmt()
+
+    # Define command.
+    # Meaning of flags:
+    # y: Answer yest to all prompts thrown by download_eggnog_data.py
```
misialq, line 27:
```suggestion
    # y: Answer yes to all prompts thrown by download_eggnog_data.py
```

q2_moshpit/plugin_setup.py
```diff
@@ -492,17 +492,43 @@
...
+    outputs=[("diamond_db", ReferenceDB[Diamond])],
+    output_descriptions={
+        "diamond_db": "Complete Diamond reference database."
+    },
+    name="Fetch the complete Diamond database necessary to run the "
+         "eggnog-diamond-search action.",
+    description="Downloads Diamond reference database.  "
+                "This action downloads 1 file (ref_db.dmnd). "
+                "At least 18 Gb of storage space is "
+                "required to run this action."
```
misialq, line 528:
```suggestion
                "required to run this action. "
```

misialq (approved):
That looks great, thanks @Sann5! Just two cosmetic changes below - ready to merge otherwise 🚀

PR #106, opened 2023-12-11
FIX: remove duplicated action registration

misialq:
This PR removes a duplicate entry for the `fetch-kaiju-db` action from `plugin_setupy.py`.


Sann5 (approved):
Looks good to me :)

PR #125, opened 2024-01-25
ENH: allow SampleData[MAGs] as input to `eggnog-diamond-search` and `classify-kraken2`

### LINKED ISSUE #110: ENH: allow `SampleData[MAGs]` as input to the `eggnog-diamond-search` action
Currently, it is possible to annotate either `contigs` or `dereplicated MAGs` - let's also allow the MAGs which have not undergone dereplication (so straight after binning), just so that annotation can be performed on any step of the pipeline.

### LINKED ISSUE #112: ENH: allow `SampleData[MAGs]` as input to the `classify-kraken2` action
For consistency, let's add `SampleData[MAGs]` as allowed input to `classify-kraken2` so that classification can be performed on every step of the pipeline.

misialq:
Closes #110.
Closes #112. 
Depends on https://github.com/qiime2/q2-types/pull/311. 

Some sample MAGs can be found [here](https://polybox.ethz.ch/index.php/s/ZbErdD6J8HpupI9/download) and dereplicated MAGs [here](https://polybox.ethz.ch/index.php/s/QAZ1Cm1XqOkl20b/download).


misialq:
Hey @gregcaporaso, do you think someone from your team could have a look/test this (particularly the Kraken2 part)? 🙏

q2_moshpit/plugin_setup.py
```diff
@@ -221,19 +227,24 @@
...
+    SampleData[Kraken2Reports % Properties('reads')]:
         SampleData[Kraken2Reports % Properties('reads')],
-    ),
-    SampleData[Kraken2Reports % Properties('contigs')]: (
+    SampleData[Kraken2Reports % Properties('contigs')]:
         SampleData[Kraken2Reports % Properties('contigs')],
-    )
+    SampleData[Kraken2Reports % Properties('mags')]:
+        SampleData[Kraken2Reports % Properties('mags')],
 })
```
misialq, line 244: Hey @ebolyen, does this look correct? This was the only way to make the TypeMap work after the addition of the "mags" property. I am not exactly sure I understand why I'm otherwise getting the error telling me the options are ambiguous...

Sann5: Just fyi this should also fail **from the command line** if I'm right about qiime2/qiime2#784

q2_moshpit/plugin_setup.py
```diff
@@ -242,19 +253,24 @@
...
+        SampleData[Kraken2Outputs % Properties('contigs', 'mags')],
+    SampleData[Kraken2Outputs % Properties('reads')]:
         SampleData[Kraken2Outputs % Properties('reads')],
-    ),
-    SampleData[Kraken2Outputs % Properties('contigs')]: (
+    SampleData[Kraken2Outputs % Properties('contigs')]:
         SampleData[Kraken2Outputs % Properties('contigs')],
-    )
+    SampleData[Kraken2Outputs % Properties('mags')]:
+        SampleData[Kraken2Outputs % Properties('mags')],
```
Sann5, line 269: Same applies here qiime2/qiime2#748

q2_moshpit/plugin_setup.py
Sann5: I think the reason why this might not have failed in the past was that the collate actions were called from within the `classify-kraken2` action, and never by themselves (e.g. `qiime moshpit collate-kraken2-reports`)

ebolyen: Hey @misialq, yes this is correct. The reason this happens and looks like a powerset sans-null, is that the properties are a kind of intersection type. It is simultaneously reads, contigs, and mags. Without the more specific branches at the top (which short-circuit) QIIME 2 reasons that a Kraken2Reports of `reads`, `contigs`, and `mags` might exist, but it would apply equally to a branch `reads`, `contigs` OR `reads`, `mags`. So it mentions there's ambiguity in the branches and hopes you fix it exactly like you did.

ebolyen: It would be nice to have a short-hand for this situation someday, because it does look pretty silly I will admit.

ebolyen: Actually, you could use `TypeMatch([...])` here instead and list only the `Properties(...)` (you'll still need the silly powerset), and then you can just use the type-variable as the property for the input and output. That'll be a bit more manageable.

ebolyen: It occurs to me that the `TypeMatch` of only properties (instead of `TypeMap` of full expressions) would also sidestep the issue @Sann5 found in the framework.

q2_moshpit/helpers/helpers.py
Sann5, line 1: Hey Michal, this action I rewrote in #137 , just fyi

Sann5: I have a good reason for having made 4 actions instead of 2 just in case you were wondering.

q2_moshpit/eggnog/tests/test_method.py
```diff
@@ -72,6 +77,31 @@ def test_good_small_search_mags(self):
...
+            },
+            index=['194b1aca-9373-4298-ba5c-05b382d1f553',
+                   '8f40a3bc-14f0-426b-b2ba-7fb3b1cd0c02',
+                   'c6bd5123-35cf-473a-bebf-85cf544bcd48',
+                   'e1ba1d20-b466-4fef-ae19-6bf9c5c63d6f']
+        )
+        exp.columns.name = 'sseqid'
+
+        pdt.assert_frame_equal(obs, exp)
+
```
Sann5, line 104: Is there a reason why here and in the other tests we actually run the search (i.e. call `_eggnog_diamond_search` without patching anything)? Would it not be preferable to patch the` subprocess.run` calls?

misialq: Hey @Sann5, yeah... This is intentional - we had discussed a while ago that for Kraken and Eggnog we'll include some tests that actually run the entire action on a minimal dataset to be able to detect potential regressions in case anything changes within those tools.

q2_moshpit/kraken2/tests/test_classification.py
```diff
@@ -781,5 +913,140 @@ def test_mags(self):
...
+        self.reports.validate()
+        self.outputs.validate()
+
+    def test_mags(self):
+        for path, df in self.output_views:
+            mag_id = str(path).rsplit('.output.txt')[0]
+            sample_id = self.uuid_to_sample[mag_id]
+
+            # the expected number of records are in the output
+            self.assertGreater(len(df), 1)
```
Sann5, line 964: Here and below, why is the `assertGreater` used instead of `assertEqual`? Do you expect that as Kraken updates are introduced the length of the output tables can increase?

Sann5 (commented):
Hey Michal! I all looks good to me. I already made some stand alone comments. As general suggestion I think I would have prefered if this was two PRs (one per action), but I imagine maybe you had a reason for doing them together.

✅ I tries the actions with some `SampleData[MAGs]` artifacts and it all works smoothly.
⚠️ All tests pass individually but when I run the whole test suite together I get error in ~12 tests for having too many files open. I think this is rather a limitation of my personal computer since this does not come up in the CI (or your own machine I suppose). Even so, this might prove annoying in the future for developers with weak computers like mine.

q2_moshpit/kraken2/tests/test_classification.py
misialq: That is a good question and it's also something I need to investigate myself 😅 I copied and modified this test as it had been written before. Will look into it ;)

misialq:
Hey @Sann5, thanks for your review! The behaviour you described, though, seems very strange - I never saw any of that when running the tests locally. Also, I don't see how this should happen as all the tests should be running one-by-one and things should get cleaned up between those runs. Could you provide more details as to that behaviour?:
- is it always the same set of tests?
- which files would those tests be using?
- how exactly are you running those?

Thanks! I'll merge this regardless as this should not be related to that PR but it would be nice if you could share some more details 🙏

PR #134, opened 2024-02-14
ENH: allow fetching 16S DBs by `build-kraken-db`

### LINKED ISSUE #92: ENH: enable fetching of kraken2's 16S databases
The `build-kraken-db` action does not yet allow fetching 16S databases, even though they are available on 
https://benlangmead.github.io/aws-indexes/k2.

We should add those to the list of options under the `--p-collection` flag, so that users interested in classification of 16S can use that action too.

DorielaGrabocka:
Added support for 4 16S databases.

Closes #92.


q2_moshpit/kraken2/database.py
```diff
@@ -220,6 +229,24 @@ def _fetch_db_collection(collection: str, tmp_dir: str):
...
+            elif (any(s16_db in item_path.lower() for s16_db in S16_DBS)
+                  and os.path.isdir(item_path)):
+                # move all files in the tmp_dir
+                extensions = ['.kmer_distrib', '.k2d', '.md', '.map']
+                for file in os.listdir(item_path):
+                    if any(file.endswith(ext) for ext in extensions):
+                        shutil.move(os.path.join(item_path, file),
+                                    os.path.join(tmp_dir, file))
+                # remove the extracted folder
+                shutil.rmtree(item_path)
```
misialq, line 248: This is overly complicated and not really necessary - since after extraction everything is wrapped into one additional directory, you just need to move all of its contents to the tmp_dir - for example, like this:
```suggestion
        extr_db = [
            x for x in os.listdir(tmp_dir) if not x.endswith(".tgz")
        ][0]
        src = os.path.join(tmp_dir, extr_db)
        for file in os.listdir(src):
            shutil.move(os.path.join(src, file), tmp_dir)
```

q2_moshpit/kraken2/tests/test_database.py
```diff
@@ -307,6 +327,39 @@ def test_fetch_db_collection_success(
...
+        mock_requests_get.has_calls([
+            call(S3_COLLECTIONS_URL),
+            call(f"{S3_COLLECTIONS_URL}/kraken/16S_Greengenes13.5.tgz",
+                 stream=True)
+        ])
+        mock_tarfile_open.assert_called_once_with(
+            "/tmp/16S_Greengenes13.5.tgz", "r:gz"
+        )
+        mock_find.assert_called_once_with("greengenes", ANY)
+        mock_tqdm.assert_not_called()
```
misialq, line 361: I think we should also have some assertions that will check your new `if` statement inside the `_fetch_db_collection` function. Or perhaps we could even have some mock files and test this part without mocking any function calls - you can try and see what's easier.

misialq (changes requested):
Hey @DorielaGrabocka, I tried it and all seems to work - see some suggestions below though 😄

.gitignore
```diff
@@ -1,143 +1,143 @@
...
-.DS_Store
-
-# VS code settings
-.vscode
-
-# Ignore notebooks
-**/*.ipynb
-
-# Ignore parsl dir
+# Byte-compiled / optimized / DLL files
```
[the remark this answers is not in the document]
DorielaGrabocka: This is a bit weird, locally I cannot see any changes. It does not detect anything. maybe after my commit it will be ok?

q2_moshpit/kraken2/tests/test_database.py
DorielaGrabocka: I completely rewrote the tests for the 16S sequences and with this, I think the if block is tested for all 4 16S database types, however since I am still a beginner in testing you can let me know if the new tests do not do what you suggest here.😊

DorielaGrabocka: Now we have tests for the method as well, as we discussed on Monday.

misialq (approved):
Hey @DorielaGrabocka, that looks good! I gave it a final try and all worked as expected. Please just see one small comment about that gitignore file. Otherwise, feel free to merge afterwards :)

PR #142, opened 2024-02-26
BUG: evaluate-busco brakes when input mags > ~1,300

### LINKED ISSUE #140: BUG: `evaluate-busco` brakes when input mags > ~1,300
The actual problem is vega, the plotting engine. [See this post.](https://altair-viz.github.io/user_guide/large_datasets.html)

Error message:
```
altair.utils.data.MaxRowsError: The number of rows in your dataset is greater than the maximum allowed (5000).
Try enabling the VegaFusion data transformer which raises this limit by pre-evaluating data
transformations in Python.
    >> import altair as alt
    >> alt.data_transformers.enable("vegafusion")
Or, see https://altair-viz.github.io/user_guide/large_datasets.html for additional information
on how to plot large datasets.
Plugin error from moshpit:
  The number of rows in your dataset is greater than the maximum allowed (5000).
  Try enabling the VegaFusion data transformer which raises this limit by pre-evaluating data
  transformations in Python.
      >> import altair as alt
      >> alt.data_transformers.enable("vegafusion")
  Or, see https://altair-viz.github.io/user_guide/large_datasets.html for additional information
  on how to plot large datasets.
```

Sann5:
### What's new
- the plotting library `vega` has a limit on the size of the plots that can be computed. 
- a user was trying to plot something really big and therefore the `evaluate-busco` action was erroring out. 
- this PR fixes that such that even when the dataset is really big a visualization is produced.
- even though we can avoid error in the action there is no going around the limit on the size of the plot. If one tries to plot more than ~1,300 MAGs the output will be an error image instead of the plot. 
- as a workaround to be able to work with datasets containing more than ~1,300 MAGs, we added a search box function to the visualization such that users can display an arbitrary subset of the samples. As long as this subset is bellow the allowed limit then a plot will be produced. 
- Closes #140 

### Run it locally 
1. Assuming you already have a local copy of `q2-moshpit` that is installed in your virtual environment. 
```bash
cd q2-moshpit
gh pr checkout 142
```

2. Let's get you some data to play with: 
```bash
cd <download here>
wget https://polybox.ethz.ch/index.php/s/slGyW1uQbyxmyb9/download -O mags.qza
```

3. Test it out!
```bash
qiime moshpit evaluate-busco --i-bins mags.qza --verbose --p-lineage-dataset bacteria_odb10 --p-cpu 6 --o-visualization mags.qzv 
```

```bash
qiime tools view mags.qzv 
```

### Running the tests
```bash
pytest -W ignore -vv --pyargs q2_moshpit
```


misialq (approved):
Hey @Sann5, thanks for fixing that! I changed the code slightly to remove the dependency on the `vegatransformer` - with this present I would constantly get errors that some packages are missing in the environment. Besides, this was needed to display more than 5k rows, right? So I guess we don't need it anyway, unless you kept it there on purpose? In any case, I'm merging this one now and we can see later whether we should think about that transformer again.

Sann5:
I'm not super sure @Michal but think the transformer might be needed. I can't remember if the visualization broke earlier (with a smaller dataset) or rather if it was needed for the search box function to work correctly. But if you tried this out with a large dataset and it worked then we are good to go :).

PR #145, opened 2024-03-05
ENH: parallelize `eggnong-annotate`

misialq:
Adds support for parallelization of the `eggnog-annotate` action.


misialq:
Exceptionally, I'm merging this without review - it was tested with a small dataset and I'm going to run it on our large one now, so worst case we can open more issues in case something is on fire.

PR #148, opened 2024-03-14
ENH: group BUSCO stats into collapsible sections for better display

### LINKED ISSUE #140: BUG: `evaluate-busco` brakes when input mags > ~1,300

misialq:
This PR addresses the problem of displaying BUSCO results for hundreds and thousands of MAGs. The problem with the previous solution was its reliance on vegafusion transformer which caused some installation issues in our distro. This PR tries to approach the problem differently:
- split the entire DataFrame with all the results into smaller partitions, each containing max. 100 MAGs
- display plots per partition
- place the plots in collapsible Bootstrap accordion to not overwhelm the user with all the plots
- allow to pick a specific sample from a dropdown and identify the section of the accordion where it sits so that the user doesn't need to uncollapse everything

More to come...

Closes #140.


q2_moshpit/assets/busco/detailed_view.html
```diff
@@ -0,0 +1,258 @@
...
+          <a href="https://github.com/BioInfoTools/BBMap/blob/master/sh/stats.sh">
+            source code and documentation 
+          </a>
+          of stats.sh for more information.
+        </p>
+
+        <div style="align-items: center; display: flex">
+          <span class="header-inline">Downloads:</span>
+          <div class="'col-lg-4">
+            <div aria-label="Basic outlined example" class="btn-group" role="group">
```
Sann5, line 50: This button is repeated in the table tab. Is that intended?

q2_moshpit/assets/busco/detailed_view.html
```diff
@@ -0,0 +1,258 @@
...
+<script crossorigin="anonymous" integrity="sha256-9SEPo+fwJFpMUet/KACSwO+Z/dKMReF9q4zFhU/fT9M=" src="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/js/bootstrap.bundle.min.js"></script>
+
+<br>
+<div class="row row-cols-1 row-cols-md-2 g-4">
+  <div class="col-lg-12">
+    <div class="card mt-3 h-10">
+      <h5 class="card-header">Plot description</h5>
+      <div class="card-body">
+        <p>
+          The left plot shows the results generated by BUSCO for <b>all bins</b> and
```
Sann5, line 27:
```suggestion
          The left plot shows the results generated by BUSCO for <b>all MAGs</b> and
```

q2_moshpit/assets/busco/detailed_view.html
```diff
@@ -0,0 +1,258 @@
...
+          of the completeness in terms of the expected gene content of a genome
+          assembly, transcriptome, or annotated gene set. The results are
+          simplified into categories of complete and single-copy, complete and
+          duplicated, fragmented, or missing BUSCOs. BUSCO completeness results
+          make sense only in the context of the biology of your organism". Visit the 
+          <a href="https://busco.ezlab.org/busco_userguide.html#interpreting-the-results">BUSCO User Guide </a>
+          for more information. 
+        </p>
+        <p>
+          The right barplot shows assembly statistics calculated for each bin using BBTools.
```
Sann5, line 38:
```suggestion
          The right barplot shows assembly statistics calculated for each MAG using BBTools.
```

q2_moshpit/assets/busco/detailed_view.html
```diff
@@ -0,0 +1,258 @@
...
+      <div class="card-body">
+        <p>
+        To prevent data overload, the samples are grouped into collapsible sections below. You can
+        expand/collapse each section by clicking on the little arrow on the right side of the section header.
+        Each section will contain a maximum of 100 bins.
+        <br><br>
+        If you want to find statistics for a specific sample, you can pick your sample from the dropdown below -
+        this will find the appropriate section and expand it.
+        <br><br>
+        You can hover over the bars to obtain information about the lineage dataset used for each bin,
```
Sann5, line 69:
```suggestion
        You can hover over the bars to obtain information about the lineage dataset used for each MAG,
```

q2_moshpit/assets/busco/detailed_view.html
```diff
@@ -0,0 +1,258 @@
...
+        and the number of genes in each BUSCO category. Additionally, you can choose the assembly statistic
+        that you wish to display from the dropdown menu included in every section.
```
Sann5, line 71:
```suggestion
        that you wish to display from the dropdown menu located at the bottom of every collapsable section.
```

q2_moshpit/assets/busco/index.html
```diff
@@ -1,138 +1,212 @@
...
+                          <th scope="row">Median</th>
+                        </tr>
+                        <tr id="rowMean">
+                          <th scope="row">Mean</th>
+                        </tr>
+                        <tr id="rowMaximum">
+                          <th scope="row">Maximum</th>
+                        </tr>
+                        <tr id="rowTotal">
+                          <th scope="row">Total count</th>
```
Sann5, line 58:
```suggestion
                          <th scope="row">Number of MAGs</th>
```
I might be overkill but maybe this is more informative. I was a little puzzled when i saw "Total count"

q2_moshpit/busco/busco.py
```diff
@@ -40,69 +123,138 @@ def evaluate_busco(
...
+    _visualize_busco = ctx.get_action("moshpit", "_visualize_busco")
+
+    (partitioned_mags, ) = partition_mags(bins, num_partitions)
+    results = []
+    for mag in partitioned_mags.values():
+        (busco_result, ) = _evaluate_busco(mag, **kwargs)
+        results.append(busco_result)
+
+    collated_results, = collate_busco_results(results)
+    visualization, = _visualize_busco(collated_results)
```
Sann5, line 258: I'm confused here because the `_visualize_busco` action seems to take 2 keyword arguments (in its function definition) yet here and in the visualizer registration in `plugin_setup.py` only one input is provided/defined. Is there something happening below the hood that I'm not aware of?

Sann5, line 258: I'm also confused because the `_visualize_busco` function does not return anything and yet here he assigns the return of the call `_visualize_busco(collated_results)` to variable ` visualization`. This also has to do with the fact that we are not really calling a function but an action that is in the end a visualizer which don't return anything?

q2_moshpit/busco/busco.py
```diff
@@ -40,69 +123,138 @@ def evaluate_busco(
...
+) -> pd.DataFrame:
     kwargs = {
-        k: v for k, v in locals().items() if k not in ["output_dir", "bins"]
+        k: v for k, v in locals().items() if k not in ["bins",]
     }
 
     # Filter out all kwargs that are None, False or 0.0
     common_args = _process_common_input_params(
         processing_func=_parse_busco_params, params=kwargs
     )
```
Sann5, line 134: This function could be placed in the main `evaluate_busco` function such that it is not repeatedly called. It would be enough to parse the argument once and then pass the parsed arguments to all calls of `_evaluate_busco`.

q2_moshpit/busco/busco.py
```diff
@@ -40,69 +123,138 @@ def evaluate_busco(
...
+    _evaluate_busco = ctx.get_action("moshpit", "_evaluate_busco")
+    partition_mags = ctx.get_action("moshpit", "partition_sample_data_mags")
+    collate_busco_results = ctx.get_action("moshpit", "collate_busco_results")
+    _visualize_busco = ctx.get_action("moshpit", "_visualize_busco")
+
+    (partitioned_mags, ) = partition_mags(bins, num_partitions)
+    results = []
+    for mag in partitioned_mags.values():
+        (busco_result, ) = _evaluate_busco(mag, **kwargs)
+        results.append(busco_result)
```
Sann5, line 255: This change would be in tandem with the change proposed in the `_evaluate_busco` function.

```suggestion
    kwargs = {
        k: v for k, v in locals().items()
        if k not in ["bins", "ctx", "num_partitions"]
    }
    
    common_args = _process_common_input_params(
        processing_func=_parse_busco_params, params=kwargs
    )

    _evaluate_busco = ctx.get_action("moshpit", "_evaluate_busco")
    partition_mags = ctx.get_action("moshpit", "partition_sample_data_mags")
    collate_busco_results = ctx.get_action("moshpit", "collate_busco_results")
    _visualize_busco = ctx.get_action("moshpit", "_visualize_busco")

    (partitioned_mags, ) = partition_mags(bins, num_partitions)
    results = []
    for mag in partitioned_mags.values():
        (busco_result, ) = _evaluate_busco(mag, common_args)
        results.append(busco_result)
```

q2_moshpit/busco/busco.py
```diff
@@ -5,21 +5,104 @@
...
+        )
+    all_summaries = _rename_columns(all_summaries)
+
+    lengths = _get_mag_lengths(bins)
+    all_summaries = all_summaries.join(lengths, on="mag_id")
+
+    return all_summaries
+
+
+def _evaluate_busco(
```
Sann5, line 105: This comment is in tandem with the code suggestion in `evaluate_busco`.

- Delete all keyword arguments except for `bins`. 
- add a keyword argument named something like `common_args`. E.g. `common_args: List[str]`
- remove lines 127-135

q2_moshpit/busco/busco.py
Sann5, line 258: Is there a reason why we unpack the return value of `collate_busco_results` and `_visualize busco` here? I can imagine its something pipeline/parsl related.

q2_moshpit/busco/busco.py
```diff
@@ -40,69 +123,138 @@ def evaluate_busco(
...
+
+    _evaluate_busco = ctx.get_action("moshpit", "_evaluate_busco")
+    partition_mags = ctx.get_action("moshpit", "partition_sample_data_mags")
+    collate_busco_results = ctx.get_action("moshpit", "collate_busco_results")
+    _visualize_busco = ctx.get_action("moshpit", "_visualize_busco")
+
+    (partitioned_mags, ) = partition_mags(bins, num_partitions)
+    results = []
+    for mag in partitioned_mags.values():
+        (busco_result, ) = _evaluate_busco(mag, **kwargs)
```
Sann5, line 254: Is there a reason why we put this into a tuple? I imagine it might be something pipeline/parsl related.

q2_moshpit/assets/busco/detailed_view.html
misialq: Yes, this is just so that one doesn't need to switch between the tabs to download the TSV.

q2_moshpit/busco/busco.py
misialq: I cannot do that - this function is registered as an action hence it requires all the arguments spelled out.

q2_moshpit/busco/busco.py
misialq: All actions return result tuples, even if they only contain a single element.

q2_moshpit/busco/busco.py
misialq: Same as above.

q2_moshpit/busco/utils.py
```diff
@@ -44,461 +49,163 @@ def _parse_busco_params(arg_key, arg_val) -> List[str]:
...
-
-
-def _run_busco(
-    output_dir: str, mags: MultiMAGSequencesDirFmt, params: List[str]
-) -> Dict[str, str]:
-    """Evaluates bins for all samples using BUSCO.
+    This function groups the DataFrame by 'sample_id' and then partitions
+    these groups into smaller DataFrames. Each partition will have a total
+    row count less than or equal to the max_rows parameter. The last group
+    in a partition can exceed the max_rows limit.
```
Sann5, line 60:
```suggestion
    This function groups the DataFrame by 'sample_id' and then partitions
    these groups into smaller DataFrames. Provided that no sample has 
    more than 'max_rows' MAGs (i.e. rows), each partition will have a total
    row count less than or equal to the 'max_rows'.
```

I think the above description is more appropriate since the first partition can also exceed the limit if, for example, the first sample has 101 MAGs/rows.

q2_moshpit/busco/busco.py
misialq: No, as I said, it is _registered_ as an action (see plugin_setup.py) and so it needs all the params defined here. Then, the pipeline fetches and runs this action, which means they have their params defined independently.

misialq: And yes, technically, you can use it as a standalone action, although it is hidden from the user (as it starts with `_`).

q2_moshpit/busco/plots_summary.py
```diff
@@ -0,0 +1,139 @@
...
+def _draw_marker_summary_histograms(data: pd.DataFrame) -> dict:
+    """
+    Draws summary histograms for the BUSCO marker results of all samples.
+
+    Returns:
+        dict: Dictionary containing the Vega spec.
+    """
+    chart = _draw_horizontal_histograms(
+        data, columns=["single", "duplicated", "fragmented", "missing"]
+    )
```
Sann5, line 61: Screenshot 2024-04-25 at 10 45 18

I think it would be useful if the domain and the range of the top-row histograms was the same for all four categories (i.e. 0-100 domain and 0-'max_counts_in_any_bin' range). This might make the top-row histogram comparison more informative. I think one can do this easily in Altair, either by [making a faceted histogram](https://altair-viz.github.io/gallery/distributions_faceted_histogram.html) or by [manually setting the scale of each chart](https://altair-viz.github.io/user_guide/customization.html#adjusting-axis-labels:~:text=To%20specify%20exact,Horsepower%3AQ%27%0A)).

q2_moshpit/busco/plots_summary.py
```diff
@@ -0,0 +1,139 @@
...
+    ).configure_axis(
+        labelFontSize=LABEL_FONT_SIZE, titleFontSize=TITLE_FONT_SIZE
+    ).configure_legend(
+        labelFontSize=LABEL_FONT_SIZE, titleFontSize=TITLE_FONT_SIZE
+    ).configure_header(
+        labelFontSize=LABEL_FONT_SIZE, titleFontSize=TITLE_FONT_SIZE
+    ).properties(
+        width=PLOT_DIM,
+        height=PLOT_DIM
+    )
```
Sann5, line 137: Screenshot 2024-04-25 at 11 38 35

In the same vein as the suggestion above, I think it would improve comparability if the domain of the per-sample histogram is the same as that of the all-samples histogram. This might be a little more annoying to implement.

q2_moshpit/busco/busco.py
Sann5: > No, as I said, it is _registered_ as an action (see plugin_setup.py) and so it needs all the params defined here. Then, the pipeline fetches and runs this action, which means they have their params defined independently.

Yes, I understand. But my point is that **you get to decide which parameters to define in the registration of `_evaluate_busco`**, whether it's the "same" parameters as in `evaluate_busco` or only one (my suggestion); a list of strings (the parsed parameters for the command). Ofc having this one-parameter setup does not make sense if one (user or developer) would like to use  `_evaluate_busco` as a standalone action.

q2_moshpit/partition/mags.py
```diff
@@ -46,7 +47,7 @@ def partition_feature_data_mags(
         for mag_fp, mag_id in _mag:
             duplicate(mag_fp, result.path / os.path.basename(mag_fp))
 
-        # If num_partitions == num_mags we will only have gone through one
+        # If num_partitions == num_samples we will only have gone through one
```
Sann5, line 50: Why `num_samples` and not `num_mags` in this coment? I thought that in `FeatureData[MAG]` artifacts we don't have samples. Besides num_mags is the name of the variable that actually gets compared in the _if_ statement below. 

```suggestion
        # If num_partitions == num_mags we will only have gone through one
```

q2_moshpit/partition/mags.py
```diff
@@ -58,6 +59,11 @@ def partition_feature_data_mags(
     return partitioned_mags
 
 
+def _chunker(seq, size):
+    # source: https://stackoverflow.com/a/434328/579416
+    return [seq[pos:pos + size] for pos in range(0, len(seq), size)]
+
+
```
Sann5, line 66: This function doesn't seem to be used anywhere.

q2_moshpit/plugin_setup.py
```diff
@@ -1122,3 +1185,10 @@
                 "classification of NGS reads.",
     citations=[citations["menzel2016"]],
 )
+
+plugin.register_semantic_types(BUSCOResults)
+plugin.register_semantic_type_to_format(
+    BUSCOResults,
+    artifact_format=BUSCOResultsDirectoryFormat)
+plugin.register_formats(BUSCOResultsFormat, BUSCOResultsDirectoryFormat)
+importlib.import_module('q2_moshpit.busco.types._transformer')
```
Sann5, line 1194: Just because this is the first time we define semantic types and formats in moshpit I will ask this... do we want to keep this here or should we move it to q2-types? I anyways was working on a semantic type for the busco database in q2-types (i think that PR is pending).

q2_moshpit/busco/busco.py
misialq: No, I do not really get to decide that. Both of these parameter sets need to be pretty much the same - I should not define the `_evaluate_busco` action with "common_args" as an allowed param as this would allow passing a list of pretty much any kinds of strings which defies the semantic validation principles. Also, this action is not exposed on the CLI but a user of the Python API can easily use it, just as I am using it in the pipeline. Hence, these parameters need to stay the same in both of those functions.

q2_moshpit/busco/plots_summary.py
misialq: Actually, the way it is shown now is intentional, after a few rounds of testing different options. The reason is that for cases where you have a plot with fewer bins than in the other plots, this subplot would become pretty much unreadable. That's why I opted for the autoscaling - this way we're sure that one can always read what's on the plot.

q2_moshpit/busco/plots_summary.py
misialq: That is a cool suggestion, thanks - I'm going to create a new issue for that (after this is merged), though, as it will require some looking into, as you noticed yourself.

q2_moshpit/plugin_setup.py
misialq: Good point - I figured I'll put it here as I really don't see a use case when you'd need the BUSCOResults type without q2-moshpit installed. Following this reasoning, though, I'm wondering now where the database type should really live 🤦

q2_moshpit/busco/types/_transformer.py
```diff
@@ -0,0 +1,40 @@
...
+    with ff.open() as fh:
+        data.to_csv(fh, sep='\t', index=False, header=True)
+    return ff
+
+
+@plugin.register_transformer
+def _3(ff: BUSCOResultsFormat) -> Metadata:
+    with ff.open() as fh:
+        df = _read_dataframe(fh)
+        return Metadata(df)
```
Sann5, line 40: Transformer nubie here! Just curious, is it necessary to call `.open()` on this `BUSCOResultsFormat` or would it suffice to do something like this?

```suggestion
@plugin.register_transformer
def _1(ff: BUSCOResultsFormat) -> pd.DataFrame:
     df = _read_dataframe(ff.path)
     return df


@plugin.register_transformer
def _2(data: pd.DataFrame) -> BUSCOResultsFormat:
    ff = BUSCOResultsFormat()
    data.to_csv(ff.path, sep='\t', index=False, header=True)
    return ff


@plugin.register_transformer
def _3(ff: BUSCOResultsFormat) -> Metadata:
     df = _read_dataframe(ff.path)
     return Metadata(df)
```

misialq: Hmmm, I don't think I tried that so I'm not 100% sure, but my guess would be probably yes? Wanna try it out and let me know? 😆

q2_moshpit/busco/types/tests/test_transformer.py
```diff
@@ -0,0 +1,66 @@
...
+    def test_dataframe_to_result_transformer(self):
+        transformer = self.get_transformer(
+            pd.DataFrame, BUSCOResultsFormat
+        )
+        df = pd.read_csv(
+            self.fp, sep='\t', header=0, index_col=0, dtype='str'
+        )
+        df.index.name = 'id'
+        obs = transformer(df)
+
```
Sann5, line 51:
```suggestion

        obs.validate()
```

q2_moshpit/busco/tests/test_busco.py
```diff
@@ -5,100 +5,190 @@
...
+            lambda x: ("visualization", )
+        ])
+        mock_ctx = MagicMock(get_action=mock_action)
+        obs = evaluate_busco(
+            ctx=mock_ctx,
+            bins=self.mags,
+            num_partitions=2
+        )
+        exp = ("collated_result", "visualization")
+        self.assertTupleEqual(obs, exp)
```
Sann5, line 194: Really cool! I didn't know this was possible :)

q2_moshpit/busco/tests/test_busco.py
```diff
@@ -5,100 +5,190 @@
...
-                params=[
-                    '--mode', 'genome',
-                    '--cpu', '1',
-                    '--contig_break', '10',
-                    '--evalue', '0.001',
-                    '--limit', '3'
-                ]
+    def _prepare_summaries(self):
+        for s in ['1', '2']:
+            os.makedirs(os.path.join(self.temp_dir.name, f"sample{s}"))
```
Sann5, line 39: This is a very convenient way of working with temp dirs 😎 . I didn't know `TestPluginBase` objects had this :)

q2_moshpit/busco/plots_summary.py
Sann5: I see. That makes sense 🙃

q2_moshpit/plugin_setup.py
Sann5: I see. Well, it definitely would speed up our development process to have the database semantic type here no? Since we are more in control of the PRs here than in q2-types? I gues that would be a point in favor of having it here.

q2_moshpit/busco/types/tests/test_transformer.py
Sann5, line 51:
```suggestion
            self.fp, sep='\t', header=0, index_col=False, dtype='str'
        )
        df.index.name = 'id'
        obs = transformer(df)
        obs.validate()
```
Setting the index col to false is necessary because otherwise `validate()` fails.

Sann5 (commented):
Hey Michal! Crazy amount of work that you put into this one 👨🏼‍💻! I really like how it looks now ✨ . Also I imagine the added information with in the index page (the histograms) can be very valuable to users 📈 . 

Most of my review I did in single comments already. This is just the conclusion message.

- ✅ All test pass in my local machine. 
- ✅ I tried the action on two different inputs and I could not find any issues with the visualizations.

q2_moshpit/busco/types/_transformer.py
Sann5: > Wanna try it out and let me know? 😆

It works! I made the changes locally and all the tests passed ✅. I also tried the action with a small input and I could not find anything wrong with the visualization.

q2_moshpit/busco/plots_summary.py
Sann5: Just to add information to this... I just noticed that for one of the inputs I used to test the action, I get histograms with only one bin. 
Screenshot 2024-04-25 at 17 30 18

This is the input artifact and the command I used if you want to try it.
[w9-mags.qza.zip](https://github.com/bokulich-lab/q2-moshpit/files/15112442/w9-mags.qza.zip)
```bash
qiime moshpit evaluate-busco --i-bins w9-mags.qza --o-visualization w9-mags.qzv --p-lineage-dataset bacteria_odb10 --verbose --o-results-table results.qza --parallel --p-num-partitions 4
```

q2_moshpit/plugin_setup.py
misialq: Yeah, sounds very sensible - let's keep the BUSCO-related one here then 👍

PR #154, opened 2024-04-12
ENH: allow `SampleData[MAGs]` as input to `predict-genes-prodigal`

### LINKED ISSUE #101: ENH: allow gene prediction from `SampleData[MAGs]`
Currently, predicting genes on dereplicated MAGs is only possible. It would be great if we could also do it on un-dereplicated MAGs.

Add `SampleData[MAGs]` as accepted input to the `predict-genes-prodigal` action.

Sann5:
### What's new
- Input to the `predict-genes-prodigal` as only allowed to be `FeatureData[MAG]`. Now it can also be `SampleData[MAG]`
- Closes #101 

```[tasklist]
### Blocked by...
- [x] merge after qiime2/q2-types#322
```

### Run it locally 
1. Check out the PR branch.

> Assuming 1) you already have a local copy of `q2-moshpit` that is installed in editable mode in your activated virtual environment and 2) the working directory is `q2-moshpit`; run the following. 

```bash
PR=154
git fetch origin pull/${PR}/head:pr-${PR}
git checkout pr-${PR}
```

2. Let's get you some data to play with: 
```bash
cd my_dir
wget https://polybox.ethz.ch/index.php/s/slGyW1uQbyxmyb9/download -O sample_data_mags.qza
```

3. Test it out!
```bash
qiime moshpit predict-genes-prodigal --i-mags sample_data_mags.qza --o-genes genes.qza --o-loci loci.qza --o-proteins proteins.qza --verbose
```


q2_moshpit/prodigal/prodigal.py
```diff
@@ -24,34 +26,56 @@ def predict_genes_prodigal(
...
+                "-a",
+                os.path.join(
+                    proteins.path, f"{prefix}{file_id}_proteins.fasta"
+                ),
+                "-d",
+                os.path.join(
+                    genes.path, f"{prefix}{file_id}_genes.fasta"
+                )
+            ])
+            run_command(cmd)
```
misialq, line 60: I think this function should be rewritten slightly to separate the filename processing logic from running the actual prodigal command. Instead of accepting a list of fasta file paths, it should take a single path, a file "id" and the prefix and construct the desired command.

q2_moshpit/prodigal/prodigal.py
```diff
@@ -24,34 +26,56 @@ def predict_genes_prodigal(
...
+
+    if isinstance(mags, MAGSequencesDirFmt):
+        # Get paths to fasta files in input dir
+        fasta_files = os.listdir(mags.path)
+        _process_fasta_files(fasta_files, '', mags.path)
```
misialq, line 65: Then here you can have a small loop which will call the prodigal function, which I described above. Also, to not mess with file names you should use the `feacture_dict` method from the `MAGSequencesDirFmt` class which will give you the mapping between mag ID and its path (https://github.com/qiime2/q2-types/blob/5eea61d956135a0d848f15e2fb2910dc72c9c092/q2_types/feature_data_mag/_format.py#L30)

q2_moshpit/prodigal/prodigal.py
```diff
@@ -24,34 +26,56 @@ def predict_genes_prodigal(
...
+
+    elif isinstance(mags, MultiMAGSequencesDirFmt):
+        # List all directories / samples
+        for sample_dir in os.listdir(mags.path):
+            sample_dir_path = os.path.join(mags.path, sample_dir)
+            if os.path.isdir(sample_dir_path):
```
misialq, line 71: Why do you need to check this? If the `listdir` above gave you the path then this check here is redundant.

q2_moshpit/prodigal/prodigal.py
```diff
@@ -24,34 +26,56 @@ def predict_genes_prodigal(
...
+                fasta_files = os.listdir(sample_dir_path)
+                print(fasta_files)
```
misialq, line 73: Please remove.

q2_moshpit/prodigal/prodigal.py
```diff
@@ -24,34 +26,56 @@ def predict_genes_prodigal(
...
+                _process_fasta_files(
```
misialq, line 74: This can then also be rewritten like the loop above, just that here you should use the `sample_dict` method from the `MultiMAGSequencesDirFmt` (https://github.com/qiime2/q2-types/blob/d24a69d649c98307851bee01b8c61a684ad864d1/q2_types/per_sample_sequences/_format.py#L608) to get all the IDs and paths.

misialq (changes requested):
Hey @Sann5, see some refactor suggestions below.

Sann5:
Hey @misialq 🙌🏼 . I did quite a few changes to the code with your suggestions in mind. Please take a look and tell me what you think. While I have you here I want to clarify something we previously discussed.

This action outputs 3 artifacts whose semantic types are `GenomeData[Loci]`, `GenomeData[Genes]`, and `GenomeData[Proteins]`. The formats associated with these semantic types have the following structure:

```python
class <name>DirectoryFormat(model.DirectoryFormat):
    <name> = model.FileCollection(r'(.*\_)?<name>[0-9]*\.<extension>$',
                                format=<format>)
```

- where `<name>` is either `loci`, `genes` or `proteins`
- `<format>` is the format of the files like FASTA of gff and `<extension>` is the associated file extension

We discussed changing the regex to something like
```regex
(.*\/)?(.*)\.<extension>$
```
bascially matching any random file name with an optional parent directory. 

Now... for the question... Right now the files that are created in the output artifacts all have this `<name>` thing in the middle. Should I already remove it (this `<name>` string)? That would mean that I have to open a PR on q2-types making these changes to the formats and then this PR would depend on that one.

misialq:
Hey @Sann5, yeah, that would be great! As discussed, there is really no need to have that <name> in the file name - the regex should pretty much be based on the extension itself as the specific formats are anyways linked to their respective semantic types so there is no need to indicate that in file names (it only makes it too restrictive).

Sann5:
Hey @misialq. I did as we discussed :). I also made the PR as dependent.

misialq (approved):
Hey @Sann5, looks good to me! I tried to run through both, sample data and feature data annotation and all worked as expected. Feel free to merge after updating the branch and resolving the conflicts ✅

PR #169, opened 2024-05-16
ENH: add actions to filter MAGs

### LINKED ISSUE #160: ENH: add an action to filter `SampleData[MAGs]` and `FeatureData[MAG]`
Similarly to what we want to have for contigs (https://github.com/bokulich-lab/q2-assembly/issues/84) we should have an action allowing filtering of MAGs based on provided metadata. (One use case would be to filter those based on BUSCOResults, another one could be to filter by ID)

```[tasklist]
### Tasks
- [ ] https://github.com/bokulich-lab/q2-moshpit/issues/165
- [ ] https://github.com/bokulich-lab/q2-moshpit/issues/166
```

misialq:
Closes #160.

Some artifacts useful for testing can be found [here](https://polybox.ethz.ch/index.php/s/PvtZhtBhB5C5Hok/download).


VinzentRisch (changes requested):
Hi @misialq  
I ran it with the provided data with no problems.
I have one suggestion. If I understand correctly the metadata is not optional and has to be provided for the filter actions. If the metadata is not given in the command, the error message is `NoneType' object has no attribute 'get_ids`, what is not very informative. I would add a check that raises an error if metadata is None, with an informative error message.
Apart from that all looks good! :)

misialq:
Hey @VinzentRisch, thanks, good catch! Actually, it does not make much sense for the metadata to be optional - otherwise, there's nothing to filter on... I just turned it into a required argument - hope that addresses your comment ;)

VinzentRisch:
Ah yeah that makes sense. 
Now I'm thinking that `where` should also be mandatory.
I saw that if `where` is None then an artifact is returned with all the original Mags.
But this does not have any application right? It just copies the artifact. So i dont think that `where` should be allowed to be None.

misialq:
No, it should not be mandatory. You can also provide a metadata file which contains the IDs you want to filter on - there is no need to provide the "where" condition in such a case (you may already have a metadata table which is pre-filtered). This is explained in the help description of `metadata` and `where` parameters.

VinzentRisch:
Ah okay that makes sense.

VinzentRisch (approved):
Looks good :)

PR #177, opened 2024-06-07
ENH: add action to estimate MAG abundance

misialq:
This PR adds an action to estimate abundances of dereplicated MAGs in the original samples. The estimation is done in the following way:
1. Reads are mapped to the dereplicated MAGs (using the `map-reads` action from q2-assembly).
2. Coverage of each MAG is calculated using the `samtools coverage` command.
3. The metric of choice (RPKM or TPM) is calculated using the coverage report (from which the read counts per MAG are obtained) and reported in the final table.

Depends on https://github.com/bokulich-lab/q2-assembly/pull/81.


misialq:
For reviewers: this action was already tested against some real data so (unless you want to try it out) feel free to focus mostly on the code review. Thanks! 🙏

ChristosMatzoros (approved):
Hey @misialq, the code looks good to me! You can merge! 🚀

PR #178, opened 2024-06-18
ENH: add action to extract a specific annotation into a feature table

misialq:
This PR introduces two new actions:
1. `extract-annotations` which can accept a `GenomeData[NOG]` artifact and extract the specified annotation (e.g. KEGG_pathways) into a new feature table
2. `multiply-tables` which can take two feature tables (with matching dimensions) and return their dot product - here the use case is calculating per-sample counts of annotation features extracted with the action 1. by getting the dot product of that table and the MAG abundance feature table which can be obtained from an action in #177.


q2_moshpit/eggnog/annotation.py
```diff
@@ -83,3 +86,100 @@ def eggnog_annotate(
...
+def _extract_brite(data: pd.DataFrame) -> pd.Series:
+    return _extract_generic(
+        data, "BRITE", lambda x: pd.Series(x.split(","))
+    )
+
+
+def _extract_caz(data: pd.DataFrame) -> pd.Series:
+    return _extract_generic(
+        data, "CAZy", lambda x: pd.Series(x.split(","))
+    )
```
ChristosMatzoros, line 145: Perhaps implementing something like the following could reduce repetition and the number of functions?

```
extraction_methods = {
    "cog": ("COG_category", lambda x: pd.Series(list(x))),
    "kegg_ko": ("KEGG_ko", lambda x: pd.Series([i[3:] for i in x.split(",")])),
    "kegg_pathway": ("KEGG_Pathway", lambda x: pd.Series([i for i in x.split(",") if i.startsWith("map")])),
    "kegg_module": ("KEGG_Module", lambda x: pd.Series(x.split(","))),
    "kegg_reaction": ("KEGG_Reaction", lambda x: pd.Series(x.split(","))),
    "brite": ("BRITE", lambda x: pd.Series(x.split(","))),
    "caz": ("CAZy", lambda x: pd.Series(x.split(",")))
}

def extract_annotations(...):
    annotation_field, transform_func = extraction_methods.get(annotation)
    if not annotation_field:
        raise NotImplementedError(f"Annotation '{annotation}' not supported.")
    ...
```

q2_moshpit/_utils.py
```diff
@@ -100,3 +103,62 @@ def get_feature_lengths(features: MAGSequencesDirFmt) -> pd.DataFrame:
...
+
+def _multiply(
+        table1: pd.DataFrame, table2: pd.DataFrame
+) -> pd.DataFrame:
+    """Calculate dot product of two tables."""
+    # if table1.shape[1] != table2.shape[0]:
+    #     raise ValueError(
+    #         f"Tables do not have compatible dimensions for dot product: "
+    #         f"{table1.shape[1]} != {table2.shape[0]}."
+    #     )
```
ChristosMatzoros, line 116: Do we need this comment?

ChristosMatzoros (changes requested):
Looks good, see the comments below.

q2_moshpit/eggnog/annotation.py
misialq: You know, my original idea was to do it exactly like this and then I didn't like it for some reason but now when I see it again I think it's much nicer this way so yeah, I'm changing this now - thanks for the suggestion!

q2_moshpit/_utils.py
misialq: Oh, no, just some left-over - I'm removing it.

ChristosMatzoros (approved):
Hey @misialq, thanks for making the changes. It looks great! 🚀

PR #181, opened 2024-06-20
TEST: include missing busco test files

Sann5:
`test_busco_feature_data.py::TestBUSCOFeatureData::test_evaluate_busco_action` and `test_busco_sample_data.py::TestBUSCOSampleData::test_evaluate_busco_action` failing because test data is missing. This PR includes such data by specifying it in the `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ setup(
             "data/*/*/*",
             "data/*/*/*/*",
             "data/*/*/*/*/*",
+            "data/*/*/*/*/*/*",
         ],
         "q2_moshpit.busco.types.tests": [
             "data/*",
--- a/setup.py
+++ b/setup.py
@@ -42,13 +42,18 @@ setup(
             'data/depth/*', 'data/maps/*', 'data/bins-small/*/*',
             'data/bins-no-uuid/*/*'
         ],
-        "q2_moshpit.busco.tests": [
-            "data/*",
-            "data/*/*",
-            "data/*/*/*",
-            "data/*/*/*/*",
-            "data/*/*/*/*/*",
-            "data/*/*/*/*/*/*",
+        "q2_moshpit.busco.tests.data": [
+            "*",
+            "summaries/*",
+            "mags/*",
+            "mags/sample1/*",
+            "mags/sample2/*",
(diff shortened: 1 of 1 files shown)


Sann5:
@misialq I made a fresh clone locally, checkout out the PR branch, and ran the tests. It seems all is good now.

misialq:
Hey @Sann5, I don't think it works - this is not something you can easily test locally - see the failing Q2 CI run. When you want to test it again after your new pushes you can just invoke the Q2 CI by commenting `/q2ci` here.

Sann5:
> Hey @Sann5, I don't think it works - this is not something you can easily test locally - see the failing Q2 CI run. When you want to test it again after your new pushes you can just invoke the Q2 CI by commenting `/q2ci` here. **Edit:** Running QIIME 2 CI...

The error in the Q2 CI is about the Python version no? Not the Busco tests. This is not the same error. But you are right I thought that these files would not be included in my fresh clone but I think they are.

misialq:
I think you're looking at the wrong action output (_our_ CI fails with Python version not the Q2 one). See here: https://github.com/bokulich-lab/q2-moshpit/actions/runs/9599409582/job/26473075711 (that's the one from my previous run)

Sann5:
So I found out that [placeholders (`*`) only work with files, not directories](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#:~:text=These%20placeholders%20are,section%20subdirectories%20below.). I find this weird becasue we have been using these placeholders for folders in other modules and all seems fine. Maybe it breaks here because there are instances of a folder with nothing but another folder inside (`busco/busco_downloads` and `lineages/lineage_1`). So I modified the setup.py to explicitly include these folders but I keep getting the error. 

@misialq could it be that the CI is not running on the right commit? It's weird that I keep getting the error even though I explicitly include these files in the setup.py. Besides, look at the commit referenced in the action. It is the commit from the last PR.

Screenshot 2024-06-21 at 10 53 42
Screenshot 2024-06-21 at 10 53 56

misialq:
Oh yeah, it certainly looks this way... Another thing to fix 🙄 I guess it looks good otherwise so let's merge this and 🤞

PR #183, opened 2024-06-21
FIX: correctly label 'Unclassfied' MAGs in the respective taxonomy

### LINKED ISSUE #179: BUG: unclassified MAGs result in empty taxonomic annotations with kraken2-to-mag-features in QIIME 2 2024.5
When attempting to generate a taxonomy table using the kraken2-to-mag-features action with the QIIME 2 2024.5 shotgun distribution, MAGs that are not successfully classified with Kraken2 are left with an empty taxonomic annotation (see first image). However, this is problematic as downstream actions in QIIME, such as taxa barplot (see error in second image), require unclassified features to be annotated at least as "d__Unclassified".

Screenshot 2024-06-18 at 16 46 05

Screenshot 2024-06-18 at 16 47 39

Here I also attach the relevant data from the inputs I used for performing kraken2-to-mag-features action and the outputted taxonomy.

[k2-to-features bug.zip](https://github.com/user-attachments/files/15901936/k2-to-features.bug.zip)

misialq:
Closes #179.

When a MAG has a mix of bacterial and eukaryotic contigs our LCA selection marks those taxonomies as NaN, while they probably should just be labeled as "d__Unclassified" (otherwise, other actions, like taxa barplot, fail as they can't handle those NaNs).

### DIFF
--- a/q2_moshpit/kraken2/select.py
+++ b/q2_moshpit/kraken2/select.py
@@ -23,6 +23,12 @@ import skbio
 RANKS = 'dkpcofgs'
 
 
+def _fill_unclassified(row):
+    if row.isnull().all():
+        row[0] = 'Unclassified'
+    return row
+
+
 def _find_lcas(taxa_list: List[pd.DataFrame], mode: str):
     """Find the least common ancestor in every DataFrame of taxa.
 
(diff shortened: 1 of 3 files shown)


misialq:
Hey @nbokulich, what do you think about this one?

nbokulich (approved):
LGTM, thanks!

PR #185, opened 2024-06-28
MAINT: make PR template available

Sann5:
Right now the PR template that we have is not being loaded when opening a new PR. I hope these changes will load every time a PR is opened.


misialq:
Let's review this once again after the October release. Some things might actually get streamlined once the distro is polished and all the dependencies are there and are not about to change anymore. This may help us come up with a slightly revised version of the template. Closing for now.

PR #187, opened 2024-07-03
PREP: updating pkg data dirs to fix metagenome prepare failures

lizgehret:
Hey @misialq, I noticed some test failures in moshpit from last weekend's [Prepare PR](https://github.com/qiime2/distributions/actions/runs/9766414380/job/26963340843?pr=287#step:9:2689) for metagenome. Looks like some package data was missing - so I've just opened this PR up on a `Prepare-` branch which will allow me to re-run an actual Prepare (PR [here](https://github.com/qiime2/distributions/pull/288)) for metagenome to confirm that this will fix the problem.


lizgehret:
closing in favor of the same changes under a `Language-` branch name (PR [here](https://github.com/bokulich-lab/q2-moshpit/pull/188)). We updated this with the python 3.9 migration, so these kinds of changes will only be picked up with a `Language-` branch (vs. a Prepare branch) - although I suspect we should fix this so both are accepted.

PR #191, opened 2024-07-22
FIX: Kraken2/Bracken feature tables/taxonomies should contain unclassified counts

misialq:
This PR changes sligthly how the feature table (and the corresponding taxonomy) is generated by Kraken2/Bracken to include counts of unclassified reads. It addresses two cases:
1. read classification + re-estimation with Bracken: the resulting feature table and taxonomy do not contain info about the unclassified reads. This is now updated to maintain this information by adding a new entry in the feature table where unclassified read counts are populated directly from the Kraken 2 reports corresponding to each sample. The taxonomy is updated by addition of an entry "d__Unclassified" (with feature ID = 0).
2. taxonomy generation after MAG classification: the resulting taxonomy skips the MAGs for which no contig could be classified. This is now updated to keep those MAGs in the taxonomy table and label them as "d__Unclassified".


misialq:
For the reviewers: we tested the code on different use cases with @pmomocabr already so please just focus on reviewing the code - thanks!

q2_moshpit/kraken2/bracken.py
```diff
@@ -111,12 +111,34 @@ def _assert_read_lens_available(
...
+) -> (pd.DataFrame, pd.Series):
+    samples = table.index.tolist()
+    for sample in samples:
+        report_fp = os.path.join(reports.path, f"{sample}.report.txt")
+        with open(report_fp, "r") as f:
+            line = f.readline()
+            if "unclassified" in line:
+                unclassified = int(line.split("\t")[1])
+                if '0' not in taxonomy.index:
+                    taxonomy.loc['0'] = "d__Unclassified"
```
ChristosMatzoros, line 127: This `if` statement can be outside the for-loop, right? It is not dependent to `sample`.

q2_moshpit/kraken2/bracken.py
```diff
@@ -111,12 +111,34 @@ def _assert_read_lens_available(
         )
 
 
+def _add_unclassified(
```
ChristosMatzoros, line 114: This needs at least one comment explaining what the function does. Maybe add some minor comments inside the function, though they are not necessary.

ChristosMatzoros (changes requested):
Looks good to me. Just have a look on my comment for a minor change. Thanks!

q2_moshpit/kraken2/bracken.py
misialq: It is dependent on samples in a sense that I need to know if there is any sample which has some "unclassified" taxonomy in them. I actually took this opportunity to change the function slightly and decouple parsing the reports from modifying the DataFrame (and incorporating your suggestion) - sorry for making you check again! I hope this code is a bit more transparent about what it's doing though.

ChristosMatzoros: Thanks for clarifying! The update makes sense.

ChristosMatzoros (approved):
Seems good @misialq! 🚀

PR #192, opened 2024-07-22
MAINT: Ortholog types updates to pair with q2-types PR#338

lizgehret:
Hey @misialq!

Me again 🤪 looks like the changes from https://github.com/qiime2/q2-types/pull/338 also require some updates in here as well. Sorry I keep making false assumptions lol but I _think_ when this is fixed we'll hopefully have a passing metagenome build 😅 otherwise I'll just keep chipping away at things, and we'll at least have a passing build before the next release!

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -27,6 +27,7 @@ requirements:
     - kraken2
     - metabat2
     - pandas
+    - parameterized
     - qiime2 {{ qiime2_epoch }}.*
     - q2-types {{ qiime2_epoch }}.*
     - q2templates {{ qiime2_epoch }}.*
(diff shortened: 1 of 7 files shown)


lizgehret:
You probably already know about this @misialq, but just in case - I looked into the test failure above and here's what I found:
Screen Shot 2024-07-22 at 2 21 48 PM

Looks like this test is failing b/c `shortuuid` wasn't imported from q2-assembly, even though that's a dep that's present in it's recipe file... I suspect this might be because the plugin deps are installed via pip (instead of through conda).

misialq:
Hey @lizgehret, thanks for the PR! I was just about to open one 😅 Let's merge this first and then I can rebase my branch - I have more changes coming (some stuff needed for future features). 

As to the shortuuid: I'm a bit confused myself as q2-assembly should be installed through conda (and with it shortuuid) but if the latest built conda package does not include that PR, maybe that's why this is failing? In any case, we should see after your prep passes :) 

Can I merge this one or should I wait for some other PR-in-progress on your side first?

lizgehret:
Thanks @misialq! Ah interesting re: q2-assembly - yeah let's see what happens after the next metagenome build. You're good to merge this PR whenever! Sorry for making any subsequent PRs you're working on more difficult 🫢

PR #200, opened 2024-08-30
ENH: parallelize Kaiju classification

### LINKED ISSUE #199: ENH: parallelize classification with Kaiju
It would be super lovely if we could introduce parsl parallelization into the `classify-kaiju` action, similarly to how it's done with Kraken and some other actions.

misialq:
Closes #199.

QIIME docs on how to use the parallelization can be found [here](https://develop.qiime2.org/en/latest/framework/how-to-guides/parallel-configuration.html).

To run you'll need any artifact with short reads and a Kaiju DB which can be fetched using the `fetch-kaiju-db` action.

Depends on #201.


ChristosMatzoros (changes requested):
Looks good to me!

VinzentRisch (approved):
Everything looks good to me!
I ran it on euler with the plasmids db and paired reads with parsl and it works without any problems.

PR #202, opened 2024-09-11
ENH: add SampleData[JoinedSequencesWithQuality] as input for classify-Kraken2

cherman2:
Hi All, 

I am currently using the metagenome distro for one of my analyses! 

I was initally strugging with my percentage unclassified from `classify-kraken2`. I found that joining my reads via vsearch and then running `classify-kraken2` was helping my classification significantly. So I thought It might be a good idea to expose this as an input to the method.  This was super easy because it is the same directory format as SampleData[SequencesWithQuality]. 

I have run this code change with my data and it runs with no issues. Feel free to close if there is a reason Joined sequences can't be used with Kraken2 that I haven't thought of!

This PR does rely on a [PR in Demux](https://github.com/qiime2/q2-demux/pull/165) so that these samples can be partitioned correctly. 

(I originally tested this PR in 2024.5 and then I merged dev into this branch so apologizes for the funky commit history). 

Let me know if you have any questions. I don't have a ton of bandwidth if this PR needs code changes but more than happy to help as much as I can! 

Depends on https://github.com/qiime2/q2-demux/pull/165


misialq (approved):
Thanks @cherman2! 🏅

PR #131, opened 2024-02-02
BUG: pin `busco` version to `5.5.0`

Sann5:
- Closes #130 
- `busco==5.6.1` requires `biopython>=1.79`. Because q2-assembly needs `biopython<=1.78` we rather pin busco to version 5.5.0. See the issue above for more info.

PR #204, opened 2024-09-12
LANG: relax busco pin for python 3.10 update

lizgehret:
Hey @misialq - I'm working on getting a solved metagenome environment for the python 3.10 update, and it looks like we'll have to bump the current version of `busco`. I did a bit of digging, and it looks like the reason for this pin came from the older `biopython` pin that was in assembly (that had to be removed for the python 3.9 update). I'm opening this paired PR so that I can test this change over in distributions, but this change won't need to go into effect until we get a solved metapackage with passing tests. So it can just be left as-is for now! I'll circle back with updates as I have them.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - altair
     - beautifulsoup4
     - bracken
-    - busco ==5.5.0
+    - busco >=5.5.0
     - diamond
     - eggnog-mapper >=2.1.10
     - gfatools


lizgehret:
PR where busco was originally pinned: #131 
PR that removed the biopython pin from the last release: https://github.com/bokulich-lab/q2-assembly/pull/80

PR open in distributions to test these changes: https://github.com/qiime2/distributions/pull/374

lizgehret:
Hey @misialq, I've got a solved environment for metagenome! Could you merge this as soon as you get a chance? 🙂

PR #206, opened 2024-09-17
Does not return Collection and List will no longer be allowed

### LINKED ISSUE #205: BUG: collate_kraken2_outputs returns list
@misialq it looks like in [this PR](https://github.com/bokulich-lab/q2-moshpit/commit/06a920b9b03d0b51fb0be934a4b6178db9772338) you changed `collate_kraken2_outputs` signature in plugin setup to say it returns a List, but I believe this is an error because as far as I can tell it only returns a single artifact. Was there a reason to register it as a List/Collection?

@ebolyen on our end, I don't think this should even be allowed by the framework, and I'm not sure how/why it worked. As I recall, you're not supposed to be allowed to register List as a return type, only Collection. I would also expect it to notice that it only got one artifact not in a list. Chloe provided me with the action.yaml of an artifact produced by this action, and it definitely does not say it is part of a list.

Oddant1:
Closes #205

### DIFF
--- a/q2_moshpit/plugin_setup.py
+++ b/q2_moshpit/plugin_setup.py
@@ -276,9 +276,8 @@ plugin.methods.register_function(
     },
     parameters={},
     outputs={
-        "collated_kraken2_outputs": List[
+        "collated_kraken2_outputs":
             SampleData[Kraken2Outputs % P_kraken_out]
-        ]
     },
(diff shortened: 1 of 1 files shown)


misialq (approved):
Hey @Oddant1, thanks for the fix - looks good to me! Unless you need this open for anything, I'll merge 🚀

PR #208, opened 2024-10-02
CI: fix mamba version and some test file paths

misialq:
This PR:
- fixes mamba version used by the setup_miniconda to 1.5.10 to prevent CI failures cause by mamba 2.0.0
- updates paths to some test files

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,7 +59,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v3
         with:
           python-version: 3.9
-          mamba-version: "*"
+          mamba-version: 1.5.10
           channels: conda-forge,defaults
           channel-priority: true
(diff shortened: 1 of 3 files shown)


VinzentRisch:
Everything looks good.
But the codecov report fails it says: Report creating failed: {"message":"Token required because branch is protected"}
I could only find this:
https://github.com/codecov/codecov-action/issues/1580

It says to create a new PR for a workaround but i guess its not that important for this one PR

VinzentRisch:
Ah I just saw you have two CI checks for coverage and the other one works.
So i guess the other one can be ignored?

misialq:
Hey @VinzentRisch, yeah, I think it's fine if only one of them works - something seems to be off, though, as the coverage for the project could not have changed that much - there were no code changes in this PR. I think I'd just merge this and address the rest in another PR, if necessary - if that's fine with you.

PR #210, opened 2024-10-14
ENH: use prodigal in eggnog-diamond-search

DorielaGrabocka:


misialq:
The coverage report is not correct as some file paths got updated so I'm just merging this PR as is.

PR #212, opened 2024-10-21
ENH: Added filter-kraken-report-output

### LINKED ISSUE #209: ENH: add action to filter Kraken reports
If some samples have only unclassified reads, passing the artifact to `estimate-bracken` action results in an error "Error: no reads found. Please check your Kraken report".

It would be useful to have an action similar to `filter-contigs` or `filter-mags` which will allow filtering reports based on criteria but also simply removal of samples without any classification results.

**Additional context**
For the linked QIIME 2 Forum discussion see [here](https://forum.qiime2.org/t/filter-kraken2-reports-e-g-before-creating-feature-tables-with-bracken/31549).

VinzentRisch:
closes #209 

New action that can filter Kraken reports and outputs by metadata and also remove empty kraken reports.

### DIFF
--- a/q2_moshpit/kraken2/__init__.py
+++ b/q2_moshpit/kraken2/__init__.py
@@ -9,8 +9,9 @@
 from .bracken import estimate_bracken
 from .database import build_kraken_db
 from .classification import classify_kraken2, _classify_kraken2
+from .filter import filter_kraken_reports_outputs
 from .select import kraken2_to_features, kraken2_to_mag_features
 
 __all__ = ['build_kraken_db', 'classify_kraken2', '_classify_kraken2',
            'estimate_bracken', 'kraken2_to_features',
-           'kraken2_to_mag_features']
+           'kraken2_to_mag_features', 'filter_kraken_reports_outputs']
--- a/q2_moshpit/kraken2/filter.py
+++ b/q2_moshpit/kraken2/filter.py
@@ -14,7 +14,8 @@ from qiime2 import Metadata
 from qiime2.util import duplicate
 
 
-def _validate_parameters(metadata, remove_empty, where, exclude_ids):
+def _validate_parameters(report_output, metadata, remove_empty,
+                         where, exclude_ids):
     if not metadata and not remove_empty:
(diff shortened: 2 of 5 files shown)


VinzentRisch:
Hi @misialq 
I have 2 questions 
1. I also added Kraken outputs to the filter actions. But it is only possible to filter by metadata and you cant remove empty outputs.
2. I saw examples of empty (unclassified) reports in the moshpit test data and my criteria for an empty report is if the report has exactly one line and if in the 6th column it says either root or unclassified. like this:
100.00	229	17	R	1	root
100.00	229	229	U	0	unclassified
Are these criteria correct or did I miss something?

misialq:
Hey @VinzentRisch, yeah, I think that is a correct assumption.

misialq (changes requested):
Hey @VinzentRisch, this overall looks good - I have a few expansion suggestions though:
- classification with Kraken 2 produces two outputs: reports and the hits - I think your action should allow filtering both, since they are pairs; maybe you can make the hits an optional input in case someone wants to filter only the reports but it should be possible to filter both at the same time
- you should add test cases to cover some of the other input combinations (for example, you don't have an example using an actual filtering query)

VinzentRisch:
Hey @misialq 
I added the support for outputs. Now outputs and reports are required. Like this I dont have to create a fake empty output artifact and the user always has both reports and outputs available anyway. 
Is it okay like this or do you see a usecase to only filter reports?

misialq (approved):
Hey @VinzentRisch, this looks great - thanks! I think it's ok if we require both - I see that @cherman2 also does that in her action in #226 so we would need to add it later anyways, most likely.

Let's wait with merging until the QIIME CI passes - probably next week.

VinzentRisch:
Okay sounds good!

misialq:
Hey @VinzentRisch, it looks like the CI is all ok now! Something occurred to me, though - should we rename this action to `filter-kraken2-results`? Now you're filtering not only reports but also the other outputs so maybe we make the name more generic?

misialq:
Thanks @VinzentRisch, nice work! 🔥

Issue #161, opened 2024-05-10
ENH: allow using BUSCOResult during MAG dereplication


misialq:
**Is your feature request related to a problem? Please describe.**
Currently, the dereplication action picks the longest MAG as the representative of the MAG cluster. We could use BUSCO results to enable picking that representative using some quality metrics and/or length.

**Describe the solution you'd like**
The `dereplicate-mags` action should accept BUSCOResults as an optional parameter. Additionally, the user should be able to indicate how they want the representative to be selected:
- if no results were provided - pick the longest one
- if BUSCOResults were provided - pick the most complete one; in case of a tie, pick the longest one

PR #213, opened 2024-10-23
ENH: Add busco-results input to `dereplicate-mags`

VinzentRisch:
solves #161 

- Adds metadata and metadata_column parameters to `dereplicate-mags`.
- If metadata and metadata_column parameters are provided the bin with the highest value in that column is chosen. If there is a tie then the longest bin is chosen.


q2_moshpit/dereplication/derep.py
```diff
@@ -257,22 +257,67 @@ def _generate_pa_table(
...
     # find similar bins, according to the threshold
     bin_clusters = _find_similar_bins_fcluster(distances, threshold)
 
-    # find the longest bin in each cluster
-    bin_lengths = _get_bin_lengths(mags)
-    longest_bins = [bin_lengths[ids].idxmax() for ids in bin_clusters]
+    # select one representative bin per cluster by BUSCO results and length
+    representative_bins = (
+            _get_representatives(mags, busco_results, bin_clusters)
+        )
```
misialq, line 317:
```suggestion
    representative_bins = _get_representatives(
        mags, busco_results, bin_clusters
    )
```

q2_moshpit/dereplication/derep.py
```diff
@@ -257,22 +257,67 @@ def _generate_pa_table(
...
+            # Get bins with the highest completeness values in cluster
+            completest_bins = (completeness_values := bin_completeness[bins])[
+                completeness_values == completeness_values.max()].index
+
+            # If there's a tie, resolve by selecting the longest bin
+            if len(completest_bins) > 1:
+                lengths_of_best_bins = bin_lengths[completest_bins]
+                representative_bins.append(lengths_of_best_bins.idxmax())
+            else:
+                representative_bins.append(completest_bins[0])
```
misialq, line 293: You could simplify this to:
```suggestion
    if busco_results is not None:
        bin_completeness = busco_results['complete']
    
        representative_bins = []
        for bins in bin_clusters:
            # Get bins with the highest completeness values in cluster
            completest_bins = bin_completeness[bins].idxmax()
    
            # If there's a tie, resolve by selecting the longest bin
            if len(completest_bins) > 1:
                representative_bins.append(bin_lengths[completest_bins].idxmax())
            else:
                representative_bins.append(completest_bins[0])
```

q2_moshpit/plugin_setup.py
```diff
@@ -418,7 +418,8 @@
     function=q2_moshpit.dereplication.dereplicate_mags,
     inputs={
         "mags": SampleData[MAGs],
-        "distance_matrix": DistanceMatrix
+        "distance_matrix": DistanceMatrix,
+        "busco_results": BUSCOResults
```
misialq, line 422: I think to be a bit more future-proof we should rather set this to be of the Metadata type and not only BUSCOResults. We would then need one more parameter which is the column name (set by default to "complete") and an additional check to make sure the column exists in the table. Before that, though, could you please check that the BUSCOResults -> q2.Metadata transformer exists? (I think it should but just in case)

misialq (changes requested):
Hey @VinzentRisch, nice work, thanks! See some suggestions below :)

q2_moshpit/dereplication/derep.py
VinzentRisch: This does not work as expected, because this will give me the first id of the bin with the highest value. But I need all ids with the highest value if there are multiple.

misialq (commented):
Hey @VinzentRisch, seems like there was a coverage failure and this time I think it's correct - could you please investigate and address accordingly? 🙏

VinzentRisch:
Hi Michal 
Now the coverage should be sorted out.
I have a question. Now the user can choose any metadata column and how its set up now the highest value in that column is used for dereplication.
Can you see any case where a user would want the lowest value in a column for dereplication.
I could add another parameter that would specify if the user wants the highest or the lowest value in the column. Do you see an application for that?

misialq:
Hey @VinzentRisch, excellent question - thanks for bringing that up! I think there may be some cases like this, although not with BUSCO. For example, CheckM has a contamination score: if a user was to present a table with those results, they may want to pick MAGs with the lowest value of contamination. So your suggestion makes sense - it would be nice if you introduce one more param and maybe set it to 'max' by default. 🚀

misialq (approved):
Looks good, thanks @VinzentRisch!

PR #218, opened 2024-11-19
BUG: fix duplicated genus in _combine_ncbi_trees

cherman2:
The solves a specific issue in scikit-bio where TreeNode.find will exit the subtree defined by a node and find an unrelated previously cached node.

This implementation uses the same algorithm as before but does not use skbio.find(). skbio's find implementation in skbio v0.6.0 was adding node cache to all nodes in the tree and never filtered the cache by subtree. Additionally these cached nodes were using taxonomic rank labels and not taxon ids and therefore were not unique allowing find to escape the subtree when a sibling was already present. 

This does not impact existing tables because it will always produce a KeyError.

This error seems to be very uncommon and a very specific collection of edge cases trigger it. 

1. There are duplicated genus names in the collection of trees
2. the first tree that is parsed has a sister clade of one of duplicated genus


ebolyen:
Coverage should be ignored. The missed branch is, in principle, impossible to hit. We were hoping coveragerc would exclude `raise AssertionError` lines/branches

misialq (approved):
Hey @cherman2, thanks for the PR! I figured I'd review this one before coming to Flagstaff again 😆 LGTM! 🏅

misialq:
Unless there's anything else you'd want to update here, I'll merge it :)

cherman2:
I think its good to merge!
thanks @misialq

PR #226, opened 2025-01-06
NEW: filter_kraken2_classifications action

cherman2:
Hi Bok Lab! 

I recently was taking to a collaborator who's area of expertise is metagenomic contaminations/spurious hits. They suggested that I should be preforming low abundance filtering on my data to filter out spurious hits. They suggested that the kraken reports is the best place to  filter, so that bracken isn't estimating taxons that were spuriously assigned.  

@colinvwood and I develped a methods to allow for filtering the kraken reports by abundance!  This method also filters the output files so that the reports and outputs are not out of sync. 

Just as a side note: I used this code on a dataset of mine and found that filtering at 0.0001 (which means the taxon has to have roughly 15-60 hits or it gets discarded) retains my diversity signal but minizes the low abundance feature overlap that was happening in my samples! 

Let me know if y'all have any questions!


q2_moshpit/plugin_setup.py
```diff
@@ -1802,6 +1802,28 @@
...
+    },
+    parameters={
+        'abundance_threshold': Float % Range(0, 100, inclusive_end=True)},
+    outputs=[('filtered_reports', SampleData[TMR]),
+             ('filtered_outputs', SampleData[TMO])],
+    input_descriptions={},
+    parameter_descriptions={},
+    output_descriptions={},
+    name='Filter Kraken2 Classifications by Abundance',
+    description='...',
```
misialq, line 1823: Should we just put here the text from the "name" field or did you want to keep the `...`?

q2_moshpit/plugin_setup.py
```diff
@@ -1802,6 +1802,28 @@
...
+        'reports': SampleData[TMR],
+        'outputs': SampleData[TMO]
+    },
+    parameters={
+        'abundance_threshold': Float % Range(0, 100, inclusive_end=True)},
+    outputs=[('filtered_reports', SampleData[TMR]),
+             ('filtered_outputs', SampleData[TMO])],
+    input_descriptions={},
+    parameter_descriptions={},
+    output_descriptions={},
```
misialq, line 1821: Please provide these missing descriptions 🙏

misialq (changes requested):
Hey @cherman2, this looks great - thanks! 🏅 

Please see the very minor cosmetic comment below.

nbokulich:
hey @cherman2 @misialq is this PR slightly overlapping with this one?
https://github.com/bokulich-lab/q2-annotate/pull/212

I think that ideally these actions should be merged into one so that we do not have different kraken report filtering actions. I like that the reports and outputs are filtered in tandem with this action.

PR #232, opened 2025-01-16
FIX: make the estimate-bracken action unavailable on OSX

misialq:
This PR removes the `estimate-bracken` action registration from plugin_setup.py in case the detected OS is OSX. The reason for this behaviour is the unavailability of Bracken on OSX platform - we don't want the users to run into weird erros.

More details in [this](https://forum.qiime2.org/t/bracken-command-not-found/32164/12) forum post.


nbokulich (approved):
LGTM, confirm that this disables that action on macOS (intel processor)

PR #236, opened 2025-02-04
ENH: expand predict-genes-prodigal to take SampleData[Contigs] as input

### LINKED ISSUE #235: ENH: Expand predict-genes-prodigal to take SampleData[Contigs] as input.
Expand predict-genes-prodigal to take SampleData[Contigs] as input. So far it only takes MAGs

VinzentRisch:
closes #235 

Expands `predict-genes-prodigal` to take SampleData[Contigs] as input.


misialq (changes requested):
LGTM overall, just a small comment below. Let's wait for @pmomocabr to test it out before merging ⏳

In the meantime, could you please investigate what happened in the CI tests?

misialq:
Also, could you please review the parameter/input/output descriptions in the plugin_setup.py? It seems like some of those require updating too 🙏

q2_annotate/plugin_setup.py
```diff
@@ -1234,13 +1234,13 @@
...
-                 "the predicted genes.",
-        'proteins': "Fasta files (one per MAG) with the protein translation "
-                    "of the predicted genes."
+        'loci': "Gene coordinates files (one per MAG or contig) listing the "
+                "location of each predicted gene as well as some additional "
+                "scoring information. ",
+        'genes': "Fasta files (one per MAG or contig) with the nucleotide "
+                 "sequences of the predicted genes.",
+        'proteins': "Fasta files (one per MAG or contig) with the protein "
+                    "translation of the predicted genes."
```
misialq, line 1243: This is incorrect - it should be "one per sample" not "per contig", right?

VinzentRisch: Ah yes true.

VinzentRisch:
So I had a look at the CI failure. It only happens in the macos build but locally the tests run fine.
There is an assertation that os.remove is called once a specific path.
The first call is the correct one:
call('/var/folders/db/x5_t5x655c9_rftmy0wp2t2w0000gn/T/qiime2/runner/processes/26598-1738857143.653269@runner/tmp/q2-OutPath-51u5bu3m/taxdmp.zip'),
The the 4 calls after that look like this:
 call(PosixPath('/var/folders/db/x5_t5x655c9_rftmy0wp2t2w0000gn/T/qiime2/runner/processes/26598-1738857143.653269@runner/a565edc1-0bf4-4357-80db-90bef9e8ec63.935881698569870148')),
 
 So it looks like MacOS CI automatically deletes temporary files and is calling os.remove to do so. 
 
 To solve this I will change the assertion to: `mock_os_rm.assert_any_call(zip_path)`

misialq (approved):
LGTM! 🔥

PR #237, opened 2025-02-04
BUG: Fixes Bug in K22F that is caused by infra-clades as terminal taxons 

cherman2:
BUG: fixes a bug in Kraken2-to-features that caused an error if an infra-clade was the terminal taxon. 

If Bracken is run before kraken2-to-features, this is not an issue becuase bracken would estimate down to the species, so this bug only occurs if Kraken2-to-features is run independently. 

Co-authored-by: @colinvwood


ebolyen (approved):
Thanks @cherman2 and @colinvwood!

PR #246, opened 2025-03-26
FIX: Allow MAGs with < 100% "unclassified" classification

lina-kim:
Currently, `kraken2_to_mag_features()` requires any unclassified MAGs to have a coverage fraction of 100.0 labeled `unclassified` by Kraken2. Consequently, execution breaks if contigs within the MAG are labeled both `unclassified` and `root`. The label `root` (alone) should be considered similarly to `unclassified`: Both labels are treated equivalently [in following code segments](https://github.com/bokulich-lab/q2-annotate/blob/main/q2_annotate/kraken2/select.py#L185-L186) and [are regarded similarly](https://github.com/DerrickWood/kraken/issues/100) by Kraken2 authors.

Points of note:
* The test `test_kraken2_to_mag_features_missing_fraction_unclassified()` may now be unnecessary. Was this meant to check that a MAG with no 'unclassified' designation (just 100% 'root') did not show up in output taxonomy files?
* The bracken module has [a similar method `_add_unclassified()`](https://github.com/bokulich-lab/q2-annotate/blob/main/q2_annotate/kraken2/bracken.py#L115-L160), but I did not implement this change there. My main goal was to fix the bug -- which does not exist in the bracken version -- but of course we can add this here
* Small bug (?): If all MAGs in an input report (to `kraken2_to_mag_features()`) are labeled 'unclassified', execution breaks [at `_find_lcas()`](https://github.com/bokulich-lab/q2-annotate/blob/main/q2_annotate/kraken2/select.py#L48) because the input `taxa_list` to concat is empty. Is this expected behavior? I understand in this case something is terribly wrong with the data quality, but was not sure whether this should be remedied


lina-kim:
Side note: I had to revert to 6be63cebffa241e5700ecba8257872f0853596d0 to avoid some errors. After ea8cb818462a098651169f2c884b9005f76d75fe my code kept reaching the so-called impossible `AssertionError(f"{tip.name} could not be inserted")`.

misialq:
Closing this in favour of #261.

PR #251, opened 2025-03-31
Filter pangenome

### LINKED ISSUE #250: ENH: extract pangenome index construction into a separate action
Currently the `filter-reads-pangenome` action constructs the pangenome index and filters against it. It would be super-useful to create a separate action which constructs the index itself and then use it within the existing action to decouple the two parts of this process. That way, users who already have the index can just use the existing filtering action from q2-quality-control.

misialq:
Closes #250.

Additionally, this PR updates the filtering action to be in line with the recent changes to the `get-ncbi-genomes` in RESCRIPT (see below).

Depends on https://github.com/bokulich-lab/RESCRIPt/pull/218.

### DIFF
--- a/q2_annotate/filtering/__init__.py
+++ b/q2_annotate/filtering/__init__.py
@@ -7,6 +7,11 @@
 # ----------------------------------------------------------------------------
 
 from .filter_mags import filter_derep_mags, filter_mags
-from .filter_pangenome import filter_reads_pangenome
+from .filter_pangenome import filter_reads_pangenome, construct_pangenome_index
 
-__all__ = ["filter_derep_mags", "filter_mags", "filter_reads_pangenome"]
+__all__ = [
+    "filter_derep_mags",
+    "filter_mags",
+    "filter_reads_pangenome",
+    "construct_pangenome_index"
+]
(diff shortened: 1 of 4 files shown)


misialq:
Hey @VinzentRisch, yup, that's what is happening - the previous API was deprecated and that broke RESCRIPt. We cannot merge that before the next release, though, as it changes dependencies (I had to update the Python client for fetching data from NCBI). So maybe you just look through the code and test afterwards, after the release? Or, if you prefer, you can also just do it all in one go later - your call.

misialq:
Hey @VinzentRisch, I merged the PR in rescript so you're good to go with reviewing this one 🙌

VinzentRisch:
Hey @misialq 
Does it make sense to create the index on the laptop? Or is this better done on euler?

misialq:
Hey @VinzentRisch, I think it's possible to do it on a laptop but you'll need a dozen GBs free space to first fetch all the required data :)

VinzentRisch (approved):
Hi @misialq 
I ran the action on euler and everything works with no problems for me.
See some minor suggestions below.

PR #253, opened 2025-04-02
MAINT: change input and output names for consistency

VinzentRisch:
- Changed all instances of "sequences" to "seqs".
- Changed all inputs that can be different sequence types to "seqs" and all inputs that are only reads to "reads".
- Changed all instances of "bins" to "mags".
- Changed all kraken reports to "reports".
- Changed all kraken outputs to "outputs" (where sometimes called hits).
- Changed all instances of "db" to the versions "kraken2_db" and "kaiju_db".
- Changed all instances of "mapped_reads" to "alignment_maps".

### DIFF
--- a/q2_annotate/plugin_setup.py
+++ b/q2_annotate/plugin_setup.py
@@ -739,8 +739,7 @@ plugin.pipelines.register_function(
     },
     input_descriptions={
         'seqs': 'Sequences to be searched for hits.',
-        "db": "Collection of profile HMMs in binary format "
-                          "and indexed.",
+        "db": "Collection of profile HMMs in binary format and indexed.",
         "idmap": "List of protein families in `hmm_db`.",
         "seed_alignments": "Seed alignments for the protein families in "
                           "`hmm_db`."
@@ -829,8 +828,7 @@ plugin.methods.register_function(
     input_descriptions={
         'seqs': 'Sequence data of the contigs we want to search for hits.',
         'idmap': 'List of protein families in `hmm_db`.',
-        'db': 'Collection of Profile HMMs in binary format '
-                          'and indexed.',
+        'db': 'Collection of Profile HMMs in binary format and indexed.',
         'seed_alignments': 'Seed alignments for the protein families in '
                           '`hmm_db`.'
     },
--- a/q2_annotate/plugin_setup.py
+++ b/q2_annotate/plugin_setup.py
@@ -696,7 +696,7 @@ plugin.pipelines.register_function(
     },
     input_descriptions={
         'seqs': 'Sequence data of the contigs we want to '
-                     'search for hits using the Diamond Database',
+                'search for hits using the Diamond Database',
         'diamond_db': 'The filepath to an artifact containing the '
                       'Diamond database',
     },
@@ -823,8 +823,7 @@ plugin.methods.register_function(
         'db_in_memory': Bool,
     },
     input_descriptions={
-        'seqs': 'Sequence data of the contigs we want to '
-                     'search for hits.',
+        'seqs': 'Sequence data of the contigs we want to search for hits.',
         'idmap': 'List of protein families in `hmm_db`.',
         'pressed_hmm_db': 'Collection of Profile HMMs in binary format '
                           'and indexed.',
--- a/q2_annotate/plugin_setup.py
+++ b/q2_annotate/plugin_setup.py
@@ -195,7 +195,7 @@ plugin.pipelines.register_function(
     parameters={**kraken2_params, **partition_params},
     outputs=[
         ('reports', T_kraken_out_rep),
-        ('hits', T_kraken_out_hits),
+        ('outputs', T_kraken_out_hits),
     ],
     input_descriptions={
         "seqs": "Sequences to be classified. Both, single-/paired-end reads"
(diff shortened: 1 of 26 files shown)


q2_annotate/plugin_setup.py
```diff
@@ -1323,19 +1322,19 @@
 plugin.methods.register_function(
     function=q2_annotate.kaiju._classify_kaiju,
     inputs={
-        "seqs": SampleData[
+        "reads": SampleData[
```
misialq, line 1325: Let's leave this at `seqs` - we have an open issue to introdcue more input types.

q2_annotate/plugin_setup.py
```diff
@@ -396,11 +396,11 @@
 
 plugin.methods.register_function(
     function=q2_annotate.kraken2.database.inspect_kraken2_db,
-    inputs={"db": Kraken2DB},
+    inputs={"kraken2_db": Kraken2DB},
     parameters={"threads": Int % Range(1, None)},
     outputs=[("report", Kraken2DBReport)],
     input_descriptions={
-        "db": "The Kraken 2 database for which to generate the report."
+        "kraken2_db": "The Kraken 2 database for which to generate the report."
```
misialq, line 403: Actually, I would leave this at `db`. Now that I'm going through this PR I'm remembering that we discussed once to keep those names simple in cases where it's obvious what kind of input/database should be used, i.e.: if the action is `do-something-kraken2` it is obvious that the db needs to be kraken2 and the outputs will also be `kraken2`. That's why the outputs of classify-kraken for example are only called `reports` and `outputs`. In the same vein, we can leave this at `db`. Could you please look at the other actions and make sure we follow this convention everywhere? I think in teh case of kaiju this would need to be adjusted.

q2_annotate/plugin_setup.py
```diff
@@ -1323,19 +1322,19 @@
...
     parameters=kaiju_params,
     outputs=[
         ("abundances", FeatureTable[Frequency]),
         ("taxonomy", FeatureData[Taxonomy])
     ],
     input_descriptions={
-        "seqs": "Sequences to be classified.",
-        "db": "Kaiju database.",
+        "reads": "Sequences to be classified.",
+        "kaiju_db": "Kaiju database.",
```
misialq, line 1337: Let's leave at `db` as pointed out above.

q2_annotate/plugin_setup.py
```diff
@@ -1350,19 +1349,19 @@
...
     },
     parameters={**kaiju_params, **partition_params},
     outputs=[
         ("abundances", FeatureTable[Frequency]),
         ("taxonomy", FeatureData[Taxonomy])
     ],
     input_descriptions={
-        "seqs": "Sequences to be classified.",
-        "db": "Kaiju database.",
+        "reads": "Sequences to be classified.",
```
misialq, line 1363: And here (both lines)

misialq (changes requested):
Hey @VinzentRisch, I had the first look which made me realize that maybe we should be a bit less "aggressive" about that renaming - there were some reasons for having some names in place (see below). Thanks for going through all of those, though - this is looking much more consistent now!

q2_annotate/plugin_setup.py
VinzentRisch: Okay then I will rename all databases to just db except for the cases where there are multiple dbs involved.

VinzentRisch: How do you feel about this one? Should it also just be db?
        
```
    function=q2_annotate.eggnog.search_orthologs_hmmer,
    inputs={
        'seqs': SampleData[Contigs | MAGs] | FeatureData[MAG],
        'pressed_hmm_db': ProfileHMM[PressedProtein],
```

q2_annotate/plugin_setup.py
```diff
@@ -816,24 +814,21 @@
 plugin.methods.register_function(
     function=q2_annotate.eggnog._eggnog_hmmer_search,
     inputs={
-        'sequences':
+        'seqs':
             SampleData[Contigs] | SampleData[MAGs] | FeatureData[MAG],
         'idmap': EggnogHmmerIdmap,
-        'pressed_hmm_db': ProfileHMM[PressedProtein],
+        'db': ProfileHMM[PressedProtein],
```
misialq, line 820: Same as above.

q2_annotate/plugin_setup.py
```diff
@@ -816,24 +814,21 @@
...
     },
     input_descriptions={
-        'sequences': 'Sequence data of the contigs we want to '
-                     'search for hits.',
-        'idmap': 'List of protein families in `hmm_db`.',
-        'pressed_hmm_db': 'Collection of Profile HMMs in binary format '
-                          'and indexed.',
-        'seed_alignments': 'Seed alignments for the protein families in '
-                          '`hmm_db`.'
+        'seqs': 'Sequence data of the contigs we want to search for hits.',
```
misialq, line 828: Could you please update this description to say "sequences" instead of "contigs"? I think we forgot to change when adding more input types.

VinzentRisch:
@misialq 
So all outputs of databases should just be "db" as well?

misialq:
Sure, why not - the ones I pointed out are the ones that jumped at me, but if you see more then go ahead and change those too 👍

misialq (approved):
I've run most of the actions and all seems fine. Once we merge this I will then test all of those in combination with the ones from q2-assembly.

Issue #211, opened 2024-10-18
BUG: classify-kraken2 fails when trying to process empty fastq files.


mikerobeson:
**Describe the bug**
I would not necessarily consider this a BUG *per se*, but when I ran `moshpit classify-kraken2` on my data, it turned out that some of our samples did not sequence and consisted of empty fastq files. I removed the empty sample files, everything appears to be running smoothly for the moment.

**To Reproduce**
1. Run `moshpit classify-kraken2` on any empty fastq files.

**Expected behavior**
I would expect that empty files would be logged to a file and skipped during processing.

**Error output log**
>Running qiime moshpit classify-kraken2 ...
/home/.../qiime2-metagenome-2024.5/lib/python3.9/site-packages/numpy/core/fromnumeric.py:59: FutureWarning: 'DataFrame.swapaxes' is deprecated and will be removed in a future version. Please use 'DataFrame.transpose' instead.
  return bound(*args, **kwds)
Loading database information... done.
24964735 sequences (6605.33 Mbp) processed in 4064.781s (368.5 Kseq/m, 97.50 Mbp/m).
  12785302 sequences classified (51.21%)
  12179433 sequences unclassified (48.79%)
Loading database information... done.
0 sequences (0.00 Mbp) processed in 0.741s (0.0 Kseq/m, 0.00 Mbp/m).
  0 sequences classified (-nan%)
  0 sequences unclassified (-nan%)
Traceback (most recent call last):
  File "/home/.../qiime2-metagenome-2024.5/lib/python3.9/site-packages/q2_types/kraken2/_format.py", line 52, in _validate_
    df, COLUMNS = self._to_dataframe()
  File "/home/.../qiime2-metagenome-2024.5/lib/python3.9/site-packages/q2_types/kraken2/_format.py", line 39, in _to_dataframe
    df = pd.read_csv(self.path, sep='\t', header=None)
  File "/home/.../qiime2-metagenome-2024.5/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 1026, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "/home/.../qiime2-metagenome-2024.5/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 620, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "/home/.../qiime2-metagenome-2024.5/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 1620, in __init__
[...]

PR #254, opened 2025-04-03
FIX: classify-kraken2 fails when trying to process empty fastq files (Issue #211)

emmabc125:
As reported in issue #211, `moshpit classify-kraken2` fails when there are empty fastq files. I added a check within the classify_kraken2_helper function for empty fastq files. If an empty file is encountered, the sample will be skipped and analysis will continue as normal with the remaining samples. When a sample is skipped, a warning message with the sample name will be output to the console and dummy report/output files will be created.

Note: In the issue, it was mentioned that the expected behavior was that "the empty files would be logged to a file". The current version of this check outputs the skipped files to the console, not a log file.

### DIFF
--- a/q2_annotate/kraken2/classification.py
+++ b/q2_annotate/kraken2/classification.py
@@ -198,8 +198,9 @@ def classify_kraken2_helper(
             # check for empty fastq file
             try:
                 for f in fps:
-                    pd.read_csv(f,compression='gzip')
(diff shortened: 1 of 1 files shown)


misialq:
Hey @emmabc125, thank you again for your PR! I'm closing it in favour of a different solution which we decided to go for.

PR #257, opened 2025-04-09
ENH: Add completness and contamination columns to BUSCO reports

### LINKED ISSUE #252: ENH: Add "completeness" and "contamination" to BUSCO report

In their [Git repo](https://github.com/metashot/busco), BUSCO developers define completeness and contamination as:

> Completeness and contamination
> Completeness is defined as as 100 minus the fraction of missing BUSCOs and contamination as the fraction of duplicated BUSCOs:
> 
> completeness % = 100 x (1 - Missing / Total)
> 
> contamination % = 100 x Duplicated / Complete

As I understand, completeness is different from "% Complete" value in the batch_summary.txt, which I believe is used by qiime2 for visualization. 

In the "[Filter MAGs](https://bokulich-lab.github.io/moshpit-docs/chapters/02_mag_reconstruction/reconstruction.html#filter-mags)" tutorial, it is suggested:

> In this example, we filter out any MAGs with completeness below 50%

by using:

> --p-where 'complete>50'

Am I misinterpreting the metrics, or "% Complete" and "Completeness" are different metrics? When calculated from standalone BUSCO output, "% Complete" and "Completeness" (completeness % = 100 x (1 - Missing / Total)) are different.

If I am not misinterpreting the metrics above, "Completeness" and "Contamination" can be calculated from BUSCO reports and added to the final report, so it would be possible to use both metrics for filtering.

Example:

> ***** Results: *****
> 
> 	C:87.9%[S:86.3%,D:1.6%],F:4.8%,M:7.3%,n:124	   
> 	109	Complete BUSCOs (C)			   
> 	107	Complete and single-copy BUSCOs (S)	   
> 	2	Complete and duplicated BUSCOs (D)	   
> 	6	Fragmented BUSCOs (F)			   
> 	9	Missing BUSCOs (M)			   
> 	124	Total BUSCO groups searched	

"% Complete" (as in Q2) = 87.9
"Completness" (as defined on BUSCO Git; 100 x (1 - Missing / Total)) = 100 * (1 - 9/124) = 92.74
"Contamination" (as defined on BUSCO Git; 100 x Duplicated / Complete) = 100 x 2 / 109 = 1.83

So my question is if I am misinterpreting the metrics or it should be added to the plugin.

I also encountered the issue while fetching BUSCO database (Q2 2024.10), but I already found here that it is a known issue and will be fixed.

VinzentRisch:
closes #252 

- Added optional completeness and contamination columns to `BUSCOResultsFormat`.
- Added _extract_json_data function that reads a JSON result file for one MAG and extract all relevant metrics.
- Removed _get_mag_lengths function because length is provided in the JSON files.
- Changed _run_busco that it just executes the BUSCO command.
- Changed _busco_helper to do the looping over samples and MAGs with the help of the sample_dict() and feature_dict() functions. 
- Before the sample_id field was showing "data" for all feature_data outputs. Changed it to show "" for all values.


misialq (changes requested):
Hey @VinzentRisch, thanks for the quick PR! I have a high-level comment about your calculation: you are using the JSON reports instead of the results table directly because it is more "precise" due to rounding. While this is technically correct, the difference is practically meaningless - if you look a bit at the equations you should notice that if you use the fractions (or percentages) directly you would get the same result just with fewer significant digits. That means that if you take your numbers and round them to one significant digit (which I think you are not doing but maybe you should just so that the numbers in the table are consistent), you will end up with (pretty much) the same number. Another issue with this approach is that now the final results table has two sources of truth: the summarized report generated by BUSCO and the original JSON report. For consistency, we should only be using one of those so normally, I'd ask you here to rewrite the other section to just use the JSON file instead. I do think, though, that you can just calculate those numbers from the ones existing in the table we already have. What do you think?

VinzentRisch:
I mean yes the results will be "pretty much" the same, but there will be times where the significant rounded digit will be one off. And if completeness and contamination are used for filtering it means that depending on if you are using the rounded numbers or the whole values the filtering will be different. I dont really understand the problem with the two sources of truth because both sources are coming directly from BUSCO.

misialq:
Regarding the source of truth, sticking to a single report is generally more maintainable and future-proof. It helps us focus on one consistent source rather than reconciling discrepancies from two, especially since one is essentially a parsed version of the other.

Honestly, the more I think about it the more I'm convinced we should really just parse the other, original report. Let’s avoid short-term quick fixes and take a long-term view - the JSON file contains _all_ metrics we may ever need and if BUSCO ever adds anything else they’ll most likely appear there. In hindsight, this approach might have been preferable, and it’s reassuring that the change isn’t too extensive - you need to adjust how the `_collect_summaries` action gathers the metrics and you could simply call your (adjusted) new method within there.

q2_annotate/busco/busco.py
```diff
@@ -21,93 +20,71 @@
...
+            # Extract results from JSON files for one sample
+            for mag_id, mag_fp in feature_dict.items():
+                results_mag = _extract_json_data(
+                    str(tmp), mag_id, sample_id, os.path.basename(mag_fp)
+                )
 
-    lengths = _get_mag_lengths(mags)
-    all_summaries = all_summaries.join(lengths, on="mag_id")
+                # Add results df from one sample to results_all df
+                results_all = pd.concat([results_all, results_mag])
```
misialq, line 85: You should probably only collect all the dfs into a list and concatenate them once at the very end, instead of concatenating at every iteration.

q2_annotate/busco/busco.py
```diff
@@ -21,93 +20,71 @@
...
-
-    Returns:
-        dict: Dictionary where keys are sample IDs and values are the paths
-            to the `batch_summary.txt` generated by BUSCO, e.g.
-            `tmp/busco_output/<sample_id>/batch_summary.txt`.
     """
-    base_cmd = ["busco", *params]
+    # Using -f (force) to be able to write to existing tmp dir when running feature data
+    # mags.
+    base_cmd = ["busco", "-f", *params]
```
misialq, line 46: This makes me slightly anxious. Why do we need to force this now, while in the previous implementation it was not necessary?

q2_annotate/busco/types/_format.py
```diff
@@ -16,7 +16,7 @@ class BUSCOResultsFormat(model.TextFileFormat):
         "mag_id", "sample_id", "input_file", "dataset", "complete",
         "single", "duplicated", "fragmented", "missing", "n_markers",
         "scaffold_n50", "contigs_n50", "percent_gaps", "scaffolds",
-        "length"
+        "length", "completeness", "contamination"
```
misialq, line 19: I think perhaps we should make the two new columns optional, as they are not part of the orginal BUSCO results. Also, making this required is backwards-incompatible and could break things for users who already used these actions to generate results.

misialq (changes requested):
Hey @VinzentRisch, thanks for the refactor - this looks much better now! Please some some comments below. Also, please have a look at the last transformer in `_transformer.py` - it may need an adjustment following your changes.

q2_annotate/busco/busco.py
VinzentRisch: Yeah  understand why. So before there was another directory introduced called "busco_output" that adds another level of nesting. And I thought that it was unnecessary and removed it. And then saw that there is an error with feature data mags and to fix this I added this -f flag. I only now connected the dots that the "busco_output" dir was added because of this problem. So the approaches lead to the same result, but I can add the "busco_output" approach again if you think this could lead to problems.

q2_annotate/busco/utils.py
```diff
@@ -255,19 +207,61 @@ def _calculate_summary_stats(df: pd.DataFrame) -> json:
...
+    Args:
+        base_path (str): The root directory containing BUSCO outputs.
+        mag_id (str): The MAG ID.
+        sample_id (str): The sample ID.
+        file_name (str): The filename of the MAG.
+
+    Returns:
+        pd.DataFrame: A dataframe containing BUSCO results and metadata.
+    """
+    json_path = glob.glob(os.path.join(base_path, sample_id, file_name, "*.json"))[0]
```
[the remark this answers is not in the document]
VinzentRisch: No I don't think I can. The filename follows this structure: short_summary.specific.<dataset>.<output_folder>.json
So I need the <dataset> to construct the path. But that is not accessible if auto lineage is being used. Or am I missing something?

q2_annotate/busco/types/_format.py
VinzentRisch: Yes this makes sense.

q2_annotate/busco/utils.py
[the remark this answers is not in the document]
misialq: So, here's the thing. There are many things entangled together which we should disentangle. Part of this comment belongs to the `force` comment above so please just treat those together. The `_extract_json_data` method should do exactly that - just extract whatever there is in the json file (following the [single responsibility principle](https://www.pythontutorial.net/python-oop/python-single-responsibility-principle/)). Right now it:
- looks for files
- extracts data
- adds more data to the dictionary

The way this would be a bit cleaner and maintainable would be:
- before this function is called: find the path to the report file - this becomes _the only_ argument to this method as it should not care about anything else
- run the method and get your dictionary
- after you get the results: enrich the dictionary by adding the other metrics (you can drop the data fields which were needed to do that afterwards) - this can even be a separate small method (easier to test too)

This way, you will also be able to pass a fake name to the `_run_busco` method as you are adding the additional fields afterwards.

q2_annotate/busco/busco.py
misialq: Makes sense. See my comment below - if you follow the reasoning there you should not need to force anything or rewrite anything (unless I still am missing some detail...)

q2_annotate/busco/busco.py
```diff
@@ -21,93 +20,71 @@
...
 from q2_types.feature_data_mag import MAGSequencesDirFmt
 
 
-def _run_busco(
-    output_dir: str,
-    mags: Union[MultiMAGSequencesDirFmt, MAGSequencesDirFmt],
-    params: List[str]
-) -> Dict[str, str]:
-    """Evaluates mags for all samples using BUSCO.
+def _run_busco(input_dir: str, output_dir: str, sample: str, params: List[str]):
```
misialq, line 35: Can you please rename `sample` to `sample_id` for consistency with the other methods?

misialq (commented):
Hey @VinzentRisch, yup, that looks good! I still need to run it but you can go ahead and work on the tests. Thanks!

q2_annotate/plugin_setup.py
```diff
@@ -1120,6 +1121,9 @@
     "miniprot": "Use miniprot gene predictor for eukaryote runs.",
     "scaffold_composition": "Writes ACGTN content per scaffold to a file "
                             "`scaffold_composition.txt`.",
+    "add_contam_complete": "Adds completeness and contamination values to the BUSCO "
```
misialq, line 1124: I'm wondering, should we set this to True by default? It takes quite some resources to compute it all and if someone realizes too late they didn't toggle that they'd "need" to rerun everything to get those additional metrics... Or did you have some reason to disable this by default?

misialq: Also, maybe we could name it something a bit more friendly, like "additional-metrics"? You can still explain in the description what exactly is being added.

q2_annotate/busco/utils.py
```diff
@@ -283,3 +264,42 @@ def _validate_parameters(lineage_dataset, auto_lineage,
             "If 'lineage-dataset' is provided, all the parameters 'auto-lineage', "
             "'auto-lineage-euk' and 'auto-lineage-prok' must be set to False."
         )
+
+
+def _process_busco_results(add_contam_complete, results, mag_id, file_name, sample_id):
```
misialq, line 269: Could you please order these arguments in some more logical way? Like results, sample_id, mag_id, file_name, add_contam_complete? Sorry, my ocd cannot take this 🙈

q2_annotate/busco/tests/test_utils.py
```diff
@@ -366,6 +345,114 @@ def test_validate_lineage_dataset_input_warning(self):
             }
         )
 
+    def test_extract_json_data(self):
+        obs = _extract_json_data(self.get_data_path(
+            "busco_output/short_summary.specific.iridoviridae_odb10."
+            "24dee6fe-9b84-45bb-8145-de7b092533a1.fasta.json"))
+
+        exp = {
```
misialq, line 353: Isn't this just the `self.busco_results`?

misialq (commented):
I ran it with/without the the metrics flag on a sample data and feature data artifacts and all seemed ok. Just some small _final_ things below.

q2_annotate/plugin_setup.py
VinzentRisch: Ah yeah okay I will change it.

misialq (approved):
Awesome, thanks @VinzentRisch! 🚀

PR #258, opened 2025-04-09
ENH: add validation that a lineage dataset or auto lineage is specified

### LINKED ISSUE #256: BUG: Running evaluate-busco with just default parameters fails
**Describe the bug**
Running evaluate-busco with just default parameters leads to an error. 
`local variable 'autoconfig' referenced before assignment`
This happens because it has to be specified what lineage to use with `--p-lineage-dataset` or `--p-auto-lineage` has to be set to True.

**Expected behavior**
When running with default parameters, there has to be a value error thrown with a message informing the user to specify a lineage.

VinzentRisch:
closes #256 

- Adds helper function `_validate_parameters` that raises error with message if parameters lineage_dataset, auto_lineage, auto_lineage_euk and  auto_lineage_prok are all None.

### DIFF
--- a/q2_annotate/busco/utils.py
+++ b/q2_annotate/busco/utils.py
@@ -272,9 +272,14 @@ def _get_mag_lengths(bins: Union[MultiMAGSequencesDirFmt, MAGSequencesDirFmt]):
 
 def _validate_parameters(lineage_dataset, auto_lineage,
                          auto_lineage_euk, auto_lineage_prok):
-    if all(value is None for value in locals().values()):
+    if not any([lineage_dataset, auto_lineage, auto_lineage_euk, auto_lineage_prok]):
         raise ValueError(
             "At least one of these parameters must be provided/set to True: "
             "'lineage-dataset', 'auto-lineage', 'auto-lineage_euk', "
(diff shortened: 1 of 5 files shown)


q2_annotate/busco/utils.py
```diff
@@ -268,3 +268,13 @@ def _get_mag_lengths(bins: Union[MultiMAGSequencesDirFmt, MAGSequencesDirFmt]):
             seq = skbio.io.read(mag_fp, format="fasta")
             lengths[mag_id] = sum([len(s) for s in seq])
         return pd.Series(lengths, name="length")
+
+
+def _validate_parameters(lineage_dataset, auto_lineage,
+                         auto_lineage_euk, auto_lineage_prok):
+    if all(value is None for value in locals().values()):
```
misialq, line 275: Could you also have a check to test if both `lineage_dataset` and one of the `auto` options were provided? When that happens BUSCO throws an error as they are incompatible.

VinzentRisch:
Added a second check like you asked. And fixed wrong logic of first check.

misialq (approved):
Great, thanks! 👏

PR #245, opened 2025-03-25
MAINT: update the file collection attribute on kraken2 report directory formats

colinvwood:
Depends on qiime2/q2-types#356

PR #238, opened 2025-02-19
Report abundance filtering

colinvwood:
Allows kraken2 reports to be filtered on a per-node (taxon) basis by relative abundance (read frequency).

PR #260, opened 2025-04-10
IMP: allow classify_kraken2 to take multiple input artifacts, merge results for overlapping sample IDs

colinvwood:
~~needs qiime2/q2-types#356~~
~~needs #245~~

~~This PR has become somewhat entangled with #238; this should be ignored until that has been reviewed & merged, then this diff will be cleaner (sorry).~~


q2_annotate/plugin_setup.py
```diff
@@ -75,8 +75,12 @@
     'minimum_hit_groups': 'Minimum number of hit groups (overlapping '
                           'k-mers sharing the same minimizer).',
     'quick': 'Quick operation (use first hit or hits).',
-    'report_minimizer_data': 'Include number of read-minimizers per-taxon and'
-                             ' unique read-minimizers per-taxon in the repot.'
+    'report_minimizer_data': (
+        'Include number of read-minimizers per-taxon and unique '
+        'read-minimizers per-taxon in the report. If this parameter is '
+        'enabled then merging kraken2 reports with the sample sample ID from '
```
gregcaporaso, line 81:
```suggestion
        'enabled then merging kraken2 reports with the same sample ID from '
```

q2_annotate/kraken2/merge.py
```diff
@@ -0,0 +1,420 @@
...
+
+    merged: tuple[TreeNode | None, TreeNode | None] | None = None
+    while reports:
+        first = _report_df_to_tree(reports.pop().view(pd.DataFrame))
+        if merged is None:
+            second = _report_df_to_tree(reports.pop().view(pd.DataFrame))
+        else:
+            second = merged
+
+        merged = _merge_trees(first, second)
```
ebolyen, line 235: This should do the same thing with less effort:
```suggestion
    functools.reduce(_merge_trees, 
                     _report_df_to_tree(x.view(pd.DataFrame)) for x in reports)
```

q2_annotate/kraken2/merge.py
```diff
@@ -0,0 +1,420 @@
...
+    -------
+        The merged kraken2 output format.
+    '''
+    if len(outputs) == 1:
+        return outputs[0]
+
+    merged_output = Kraken2OutputFormat()
+    while outputs:
+        with (
+            open(str(merged_output), 'a') as merged_fh,
```
ebolyen, line 268: I would open this once, outside of the while, then you don't need the somewhat unusual mode `'a'` and we can keep the same file descriptor around.

q2_annotate/kraken2/merge.py
```diff
@@ -0,0 +1,420 @@
...
+
+    # merge trees (with respect to `n_frags_assigned`, `n_frags_covered`)
+    if first_tree is None and second_tree is None:
+        merged_tree = None
+    elif first_tree is None:
+        merged_tree = second_tree
+    elif second_tree is None:
+        merged_tree = first_tree
+    else:
+        for node in first_tree.levelorder():
```
gregcaporaso, line 312: Adding comments to the individual steps your taking in here (i.e., why are you adding the specific values that you're adding; why is it same to assume a node's parent from `first` is in `second`, ...) would be helpful for interpreting this section.

gregcaporaso (commented):
This is a partial review, we just ran out of time today.

q2_annotate/plugin_setup.py
```diff
@@ -1901,6 +1932,51 @@
...
+    },
+    parameters={},
+    outputs={
+        "merged_reports": KRAKEN2_REPORTS_T,
+        "merged_outputs": KRAKEN2_OUTPUTS_T
+    },
+    input_descriptions={
+        "reports": (
+            "Multiple kraken2 reports artifacts to merge on a per-sample ID "
+            "basis."
```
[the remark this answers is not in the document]
colinvwood: I think it's important to note that the merging happens only where sample IDs are the same, but I did change to wording.

q2_annotate/kraken2/merge.py
```diff
@@ -0,0 +1,420 @@
...
+
+    Raises
+    ------
+    ValueError
+        If two MAGs with the same uuid are detected.
+    ValueError
+        If two reports with the same sample ID are to be merged but the
+        minimizers columns are present in the reports.
+    '''
+    minimizers_present = _check_for_minimizers(reports[0])
```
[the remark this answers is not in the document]
colinvwood: Because this action isn't user level. Within a `classify-kraken2` run either all reports will have minimizers reported or none of them will. There is technically the possibility that someone runs the `_merge_kraken2_results` hidden action in isolation, so I can change this to check all reports for minimizers I suppose.

q2_annotate/kraken2/merge.py
```diff
@@ -0,0 +1,433 @@
...
+    ValueError
+        If more than one matching node is found.
+    '''
+    search_taxon_id = search_node._kraken_data['taxon_id']
+
+    matches = []
+    for child in children:
+        child_taxon_id = child._kraken_data['taxon_id']
+        if search_taxon_id == child_taxon_id:
+            matches.append(child)
```
ebolyen, line 392: Someday we might change this to a dict lookup, but I don't think there's usually all that many children at any given level (especially with all of the infraclades) to make the performance win obvious.

q2_annotate/kraken2/merge.py
```diff
@@ -0,0 +1,433 @@
...
+    into_node._kraken_data['n_frags_assigned'] += \
+        from_node._kraken_data['n_frags_assigned']
+
+    for child in from_node.children:
+        match = _find_node_match(child, into_node.children)
+        if match is None:
+            into_node.append(child)
+            child.parent = into_node
+        else:
+            _merge_trees_recursively(child, match)
```
ebolyen, line 359: :100:

q2_annotate/kraken2/classification.py
```diff
@@ -53,60 +53,151 @@ def _construct_output_paths(
...
+    ctx,
+    seqs,
+    db,
+    threads=1,
+    confidence=0.0,
+    minimum_base_quality=0,
+    memory_mapping=False,
+    minimum_hit_groups=2,
+    quick=False,
+    report_minimizer_data=False,
```
ebolyen, line 65: We need to check that we don't have more than one `seqs` when this is `True`.

PR #270, opened 2025-04-25
ENH: fetch-busco-db and evaluate-busco improvements

### LINKED ISSUE #266: ENH: fetch-busco-db should allow fetching single lineages
Right now, fetching the entire domain datasets is possible - resulting databases are huge and, in many cases, unnecessary. We should allow fetching individual lineage datasets in case users want to only use/keep those - this will very significantly reduce the size of the databases.

### LINKED ISSUE #267: MAINT: remove dataset property for BuscoDB semantic type
This property only overcomplicates things and is not really necessary (or useful) - we are validating lineages inside the actions themselves.

### LINKED ISSUE #268: MAINT: clean up some of the evaluate-busco parameters
There are a couple of parameters which should be removed as they are not relevant:
- config
- force
- scaffold_composition

### LINKED ISSUE #269: MAINT: update the BuscoDB database format
The directory format unnecessarily stores all the data in the "busco_downloads" subdirectory - this should be simplified. Many of the files need to become optional to accommodate #266. The semantic type will be renamed to `ReferenceDB[BUSCO]`.

misialq:
Closes #269.
Closes #268.
Closes #267.
Closes #266.

### DIFF
--- a/q2_annotate/busco/busco.py
+++ b/q2_annotate/busco/busco.py
@@ -103,16 +103,13 @@ def _evaluate_busco(
     auto_lineage_euk: bool = False,
     auto_lineage_prok: bool = False,
     cpu: int = 1,
-    config: str = None,
     contig_break: int = 10,
     evalue: float = 1e-03,
-    force: bool = False,
     limit: int = 3,
     long: bool = False,
     metaeuk_parameters: str = None,
     metaeuk_rerun_parameters: str = None,
     miniprot: bool = False,
-    scaffold_composition: bool = False,
     additional_metrics: bool = False,
 ) -> pd.DataFrame:
     kwargs = {
@@ -121,7 +118,7 @@ def _evaluate_busco(
         ]
     }
     kwargs["offline"] = True
-    kwargs["download_path"] = f"{str(db)}/busco_downloads"
+    kwargs["download_path"] = str(db)
 
     if lineage_dataset is not None:
         _validate_lineage_dataset_input(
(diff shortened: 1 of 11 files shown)


VinzentRisch:
Hey Michal
I agree that this makes the handling of the databases much easier because you dont have to download everything. But downloading specific linages assumes that you know what lineages you will need. So to use the auto-lineage functionality you still have to download all lineages what takes such a long time. 
When running BUSCO in online mode and with auto lineage it checks what lineages are needed and then only downloads those lineages (If I understood correctly). So there could be an option to run the busco action without a db and then it runs automatically in online mode. 
What are your thoughts on this?

misialq:
Hey @VinzentRisch, that's an interesting point. So here's what I think:
1. Yes, you do have to know what lineage you want but in the very least you can run against bacteria_odb12 or something like this. BUSCO has a command to list all available lineages which can easily be run within our conda environment so that users can look up first which lineage they might need/want. 
2. Running with auto-lineage takes much more time because, as you can guess, you first need to identify which one to use so, at least for us, this is anyway not the approach - we provide lineages directly. Why we decided to not allow "free" download of anything without providing a DB is that this approach is rather wasteful - every time you need to download a bunch of data (BUSCO needs to fetch some lineages to be able to find out which ones to use).
3. The approach introduced here still allows you to download all of the "large" DBs just like it was possible before. You need to pass those to the "lineages" param instead of using the flags. So all of the things that were possible before are still possible now - you just need to download the appropriate DB.

So, in summary, these changes still allow you to do everything just like before. They only add a functionality of fetching specific lineages on top of that (and simplify the arguments of that action).

VinzentRisch (approved):
Hey Michal 
Everything looks good to me. I ran it with a couple of different combinations and all works as expected.

PR #275, opened 2025-05-23
ENH: Adds contamination and completeness to the BUSCO visualisation 

### LINKED ISSUE #271: ENH: include the new BUSCO metrics in the visualization
No, it is related to the recent PR #257 which introduced the `completeness` and `contamination` metrics to the BUSCO results report. Currently, those metrics are not included in the visualization produced by the `evaluate-busco` action.

I would like those metrics to be displayed in the results tables, wherever possible and available. Additionally, it would be great if we could have an additional scatter plot showing completeness _vs._ contamination, similarly to how we do it with CheckM (https://github.com/bokulich-lab/q2-checkm/blob/99ded76bb9d0d055c7fd6842a841c8eb98feb68b/q2_checkm/plots.py#L238)

VinzentRisch:
closes #271 

- Adds the metrics contamination and completeness to the summery statistics and histograms.
- Adds new scatterplot under histograms that displays contamination vs completeness.
- Fixes Bug where histograms were not showing for feature data because the sample_id col contained nan values that are not permitted in vega specs. To solve this all nan values get converted to None before the vega spec creation.

### DIFF
--- a/q2_annotate/busco/busco.py
+++ b/q2_annotate/busco/busco.py
@@ -140,8 +140,6 @@ def _visualize_busco(output_dir: str, results: pd.DataFrame) -> None:
         index=False
     )
 
-    comp_cont = "completeness" in results.columns and "contamination" in results.columns
-
     # Outputs different df for sample and feature data
     results = _parse_df_columns(results)
     max_rows = 100


misialq (commented):
Hey @VinzentRisch, I tried your changes on a larger set of samples and the visualziation I got had no plots in it - I noticed two null values in my results table (see attached). Can you look at it? 🙏 
[busco_results.csv](https://github.com/user-attachments/files/20480774/busco_results.csv)

VinzentRisch:
Hey @misialq 
If I click on the link to the file it says not found.

VinzentRisch:
Yeah that works, thanks.
Can you send me the artifact that you used to create these results?

VinzentRisch:
Hey @misialq 
I did all the changes. I could not figure out how to get the selection working directly with the legend so I added a selection menu where you can choose the sample ID.

misialq:
Hey @VinzentRisch, thanks - that looks better! I discovered something new, though 😢 I tried it with the `FeatureData[MAG]` kind of input and while the plot is there, there are a few issues:
- when I have many MAGs, most of them get collapsed in the legend and the top ones seem to have the same color - I suspect that it's because of the color map scaling (since there are so many) but at this point I'm wondering whether this legend is in any way useful
- there is no dropdown here so if I want to select a specific MAG I cannot (not sure how important this is but perhaps would be nice to have for consistency with the per-sample selection)
- the plots on the second tab (BUSCO plots) are broken
 
Image 1
Image 2

I feel like we cannot do much about the first two "issues" - do you have any thoughts or ideas on those? Otherwise, you should at least look at what's wrong with the second tab...

VinzentRisch:
Ah and the broken busco plot is again because of NaN values in the outputs that dont have sample IDs so this was broken already, but easily fixed! :)

misialq:
Yeah, it's hard to decide 😓 I think that having the legend there does not hurt... Worst case, one cannot find a specific MAG but I doubt anyone would anyway look. For the dropdown, again, probably not necessary... But if it's not super complicated then maybe you can add just so that we solve the problem of the legend together with this? If it's too complicated then let's leave it as is.

VinzentRisch:
No it's not complicated I added the dropdown menu. But do you mean then if the dropdown menu can be deleted or what do you mean with "solve the problem of the legend together with this"

misialq:
I just meant that with the dropdown we don't care as much about the legend - you can keep it for the cases where there are not that many MAGs and in case there are too many one can just use the dropdown.

misialq (changes requested):
Hey @VinzentRisch, all the visualizations work for me (with and without the new metrics). I do have a high level comment, though: you added the new `comp_cont` argument to a lot of the functions - there really is no need for that as most (if not all) of those functions receive the `results` as another argument, meaning all the information about whether those metrics are there is _in the actual results table_. You should rather leave the function signatures as they are and adjust the logic inside the specific methods (since you are doing it anyways). This way the code will be a bit cleaner.

q2_annotate/busco/busco.py
```diff
@@ -139,6 +139,9 @@ def _visualize_busco(output_dir: str, results: pd.DataFrame) -> None:
         os.path.join(output_dir, "busco_results.csv"),
         index=False
     )
+
+    comp_cont = "completeness" in results.columns and "contamination" in results.columns
```
misialq, line 143: This assignment is redundant as you're overriding it later in the if statement on line 229.

PR #277, opened 2025-06-16
ENH: expand the BUSCO visualization by including unbinned contigs

### LINKED ISSUE #240: ENH: expand the BUSCO visualization by including unbinned contigs
There is no way in MOSHPIT to estimate what is the proportion of the unbinned contigs to all contigs in each sample. This could be displayed in the MAG quality control visualization which comes out of the `evaluate-busco` action. 

The action would need to receive a new input (unbinned contigs), estimate the count of all contigs in all MAGs (per sample), calculate the proportion of the unbinned contigs and include this new metric in the visualization (probably on the first tab, as the other ones are more detailed views).

adsapountzi:
Closes #240.


q2_annotate/busco/busco.py
```diff
@@ -285,24 +445,63 @@ def evaluate_busco(
...
+            # Get the actual sample IDs in this partition
+            sample_ids = list(mag_partition.view(MultiMAGSequencesDirFmt).sample_dict().keys())
+
+            # Create metadata for filtering unbinned contigs
+            index = pd.Index(sample_ids, name="ID")
+            metadata = Metadata(pd.DataFrame(index=index))
+            id_list = ", ".join([f"'{sid}'" for sid in sample_ids])
+            where = f"ID IN ({id_list})"
+
+            # Filter the unbinned contigs for this partition
```
misialq, line 491: You don't need the `where` statement if you only filter by IDs - it's enough to pass just the metadata.

q2_annotate/busco/busco.py
```diff
@@ -112,6 +244,7 @@ def _busco_helper(bins, common_args):
 
 def _evaluate_busco(
     bins: Union[MultiMAGSequencesDirFmt, MAGSequencesDirFmt],
+    unbinned_contigs: ContigSequencesDirFmt, ### NEW unbinned
     busco_db: BuscoDatabaseDirFmt,
```
misialq, line 248: `unbinned_contigs` should be optional, hence we need to set those to `None`

q2_annotate/busco/busco.py
```diff
@@ -150,7 +283,33 @@ def _evaluate_busco(
         processing_func=_parse_busco_params, params=kwargs
     )
 
-    return _busco_helper(bins, common_args)
+    busco_results = _busco_helper(bins, common_args)
+    unbinned_percentage = {}
```
misialq, line 287: This section needs to be conditional on `unbinned_contigs` - if the user did not input any this should be skipped.

q2_annotate/busco/busco.py
```diff
@@ -150,7 +283,33 @@ def _evaluate_busco(
...
+    unbinned_count = {}   
+    # ### NEW Calculate unbinned percentage AFTER BUSCO RESULTS 
+    #add absolute count of unbinned contigs
+    if issubclass(type(bins), MultiMAGSequencesDirFmt):
+        ### NEW Add unbinned percentage per sample
+        if "unbinned_percentage" not in busco_results.columns:
```
misialq, line 293: Why do you need this `if` (and the one under)? Is there a case where this column would already be present in the results?

q2_annotate/busco/busco.py
```diff
@@ -30,7 +30,139 @@
...
+    total_sequences = 0
+
+    for fasta_fp, dna_iterator in unbinned_contigs.sequences.iter_views(DNAIterator):
+        sequence_count = sum(1 for _ in dna_iterator)  # Count sequences
+        # print(f"{fasta_fp} contains {sequence_count} sequences")  # Debugging output
+        total_sequences += sequence_count
+
+    return total_sequences
+
+def calculate_unbinned_percentage(bins, unbinned_contigs):
```
misialq, line 142: Please add docstrings here (and in all of your other functions) - it's a bit confusing to understand that `bins` corresponds to _all_ the bins but `unbinned_contigs` corresponds to a single sample.

q2_annotate/busco/busco.py
```diff
@@ -30,7 +30,139 @@
...
+    """ Calculates the percentage of unbinned contigs """
+    #need to remove loading artifacts
+    
+    # # Load QIIME2 artifacts and get correct directory formats
+    # mags_dir = Artifact.load(mags_qza).view(MultiMAGSequencesDirFmt)  # Binned MAGs
+    # unbinned_dir = Artifact.load(unbinned_qza).view(ContigSequencesDirFmt)  # Unbinned Contigs
+
+    # Count sequences
+    binned_contigs = count_binned_contigs(bins)
```
misialq, line 151: This should probably happen before this method is called, not here. If you do it here, you need to calculate this every time you process a new sample, which is unnecessary.

q2_annotate/busco/busco.py
```diff
@@ -30,7 +30,139 @@
...
+
+    #Iterate over sequences directly using QIIME 2's DNAIterator
+    for fasta_fp, dna_iterator in bins.sequences.iter_views(DNAIterator):
+        sequence_count = sum(1 for _ in dna_iterator)  # Count sequences
+        # print(f"{fasta_fp} contains {sequence_count} sequences")  # Debugging output
+        total_sequences += sequence_count
+
+    return total_sequences
+
+def count_unbinned_contigs(unbinned_contigs) -> int:
```
misialq, line 130: This method is _identical_ to the one above - you probably can simplify your code to only use one of those two.

misialq (changes requested):
Hey @adsapountzi, thanks for opening the PR, nice work! I pushed a commit with some changes - please se what got updated by comapring it to your last commit. Also, I made a couple of comments with improvement suggestions so be sure to check those out below :)

q2_annotate/busco/busco.py
adsapountzi: Thanks for pointing that out!

q2_annotate/busco/busco.py
adsapountzi: You're right, that makes sense

q2_annotate/busco/busco.py
adsapountzi: `filter_contigs()` raises a ValueError if `metadata` is provided without a `where`, so the `where` argument seems to  be required for now.

q2_annotate/busco/busco.py
```diff
@@ -140,6 +154,24 @@ def _evaluate_busco(
     common_args = _process_common_input_params(
         processing_func=_parse_busco_params, params=kwargs
     )
+    if isinstance(mags, MultiMAGSequencesDirFmt):
+        busco_results = _busco_helper(mags, common_args, additional_metrics)
```
misialq, line 158: Why does this need to be called twice, depending on the if-statement result? You should pull it out of this block and just modify the results accordingly. Then, there should be only a single `return` statement at the very end which will return `busco_results`. This way the flow is a bit more clear.

q2_annotate/busco/busco.py
```diff
@@ -140,6 +154,24 @@ def _evaluate_busco(
...
+        busco_results["unbinned_contigs"] = pd.NA
```
misialq, line 159: The entire content of this if branch should probably just become a small, independent function. Please refactor this and add a unit test for this specific functionality. This will make the code more modular and easy to follow.

q2_annotate/busco/busco.py
```diff
@@ -140,6 +154,24 @@ def _evaluate_busco(
...
+        busco_results["unbinned_contigs_count"] = pd.NA
+        for unbinned_id, unbinned_path in unbinned_contigs.sample_dict().items():
+            binned_dir = mags.path / unbinned_id
+            binned_fasta_paths = _get_fasta_files_from_dir(binned_dir)
```
misialq, line 163: Instead of doing this you should use the `sample_dict` method available on MultiMAGSequencesDirFmt to retrieve all the required fasta files.

q2_annotate/busco/utils.py
```diff
@@ -323,3 +402,103 @@ def _process_busco_results(results, sample_id, mag_id, file_name, additional_met
...
+
+    for fp in file_paths:
+        total_sequences += sum(
+            1 for _ in skbio.io.read(str(fp), format="fasta", constructor=skbio.DNA)
+        )
+
+    return total_sequences
+
+
+def _calculate_unbinned_percentage(
```
misialq, line 473: Please add type annotation and fix the docstring - it says that the method expects DirFmts but in fact you are passing paths to it. Also, please follow the docstring format convention we are using in other places of this repository.

q2_annotate/busco/busco.py
```diff
@@ -140,6 +154,24 @@ def _evaluate_busco(
...
+            percentage, count = _calculate_unbinned_percentage(
```
misialq, line 164: This is technically fine the way it's written but I think it would be a bit neater if, instead of updating the results on every iteration, you collected all the results into a pd.DataFrame and then just merged this DataFrame witht the original results DF at the very end. This way you need to temper with the results table only once and can take care of the data formats in all the columns in the intermediate DataFrame (which most likely will just happen automatically).

q2_annotate/busco/busco.py
```diff
@@ -286,24 +331,41 @@ def evaluate_busco(
...
     collate_busco_results = ctx.get_action("annotate", "collate_busco_results")
     _visualize_busco = ctx.get_action("annotate", "_visualize_busco")
+    _filter_contigs = ctx.get_action("assembly", "filter_contigs")
 
+    # add partition cases for unbinned contigs if bins.type <= SampleData[MAGs]:
     if issubclass(mags.format, MultiMAGSequencesDirFmt):
         partition_action = "partition_sample_data_mags"
     else:
         partition_action = "partition_feature_data_mags"
+        warnings.warn(
```
misialq, line 348: Perhaps this warning should only appear if the user actually provided some contigs, otherwise it may be confusing.

q2_annotate/busco/busco.py
```diff
@@ -286,24 +331,41 @@ def evaluate_busco(
...
+        )
+
     partition_mags = ctx.get_action("types", partition_action)
 
     (partitioned_mags,) = partition_mags(mags, num_partitions)
     results = []
-    for mag in partitioned_mags.values():
-        (busco_result,) = _evaluate_busco(mag, **kwargs)
-        results.append(busco_result)
+    if mags.type <= SampleData[MAGs]:
```
misialq, line 356: Maybe you could add a short comment here what this if statement is doing?

q2_annotate/busco/utils.py
```diff
@@ -323,3 +402,103 @@ def _process_busco_results(results, sample_id, mag_id, file_name, additional_met
...
+        unbinned_contigs (ContigSequencesDirFmt): The full unbinned contigs.
+        mag_partition (MultiMAGSequencesDirFmt): One partition of MAGs.
+        _filter_contigs (Action): QIIME 2 action to filter contigs.
+
+    Returns:
+        ContigSequencesDirFmt: Filtered unbinned contigs matching the partition samples.
+    """
+    sample_ids = list(mag_partition.view(MultiMAGSequencesDirFmt).sample_dict().keys())
+    metadata = Metadata(pd.DataFrame(index=pd.Index(sample_ids, name="ID")))
+    id_list = ", ".join([f"'{sid}'" for sid in sample_ids])
```
misialq, line 436: I know we've decided to do it this way but I actually fixed this "bug" in q2-assembly so it should now be possible to pass the metadata without the "where" statement. Could you please simplify this here too? You may need to install the latest q2-assembly using pip to get my change into your conda env.

q2_annotate/busco/plots_summary.py
```diff
@@ -209,3 +209,55 @@ def _draw_selectable_summary_histograms(data: pd.DataFrame) -> dict:
...
+def _draw_selectable_unbinned_histograms(data: pd.DataFrame) -> dict:
+    """
+    Draws summary histograms for the MAG assembly metrics where users
+    can indicate which metric and for which sample they want to see.
+
+    Returns:
+        dict: Dictionary containing the Vega spec.
+    """
+    metrics = ["unbinned_contigs", "unbinned_contigs_count"]
+    labels = ["Percentage of Unbinned Contigs", "Absolute Count of Unbinned Contigs"]
```
misialq, line 223: Let's only capitalize the first word in each name".

q2_annotate/busco/utils.py
```diff
@@ -180,9 +222,12 @@ def _parse_df_columns(df: pd.DataFrame) -> pd.DataFrame:
         df (pd.DataFrame): Formatted DataFrame
     """
     cols = MARKER_COLS.copy()
+
     if not ("completeness" in df.columns and "contamination" in df.columns):
```
misialq, line 226: Can you change this part and the contents of MARKER_COLS so that you _add_ here instead of remove? IT would then be consistent with the second branch of this statement.

q2_annotate/busco/utils.py
```diff
@@ -192,10 +237,32 @@ def _parse_df_columns(df: pd.DataFrame) -> pd.DataFrame:
     for col in cols:
         df[col] = df[col].map(float)
     df["n_markers"] = df["n_markers"].map(int)
-
     return df
 
 
+# def _rename_columns(df):
```
misialq, line 243: Why is this commented out? Can we remove that?

q2_annotate/busco/utils.py
```diff
@@ -210,6 +277,12 @@ def _calculate_summary_stats(df: pd.DataFrame) -> json:
         cols.remove("completeness")
```
misialq, line 277: Please also update this statement here as I already described above.

misialq (changes requested):
Hey @adsapountzi, thanks for all the updates! I left some comments below to help you streamline the code a bit. Feel free to respond under those if you have any questions :)

Also, please have a look at the failing checks in the UI below. One of those is flake8 - you can run it locally and fix all the warnings that it shows you.

PR #278, opened 2025-06-17
ENH: add EC number to the list of extractable annotations

### LINKED ISSUE #276: ENH: allow extraction of EC numbers with extract-annotations action
The action map-eggnog creates a table of various annotations such as GO terms, KEGG_ko terms or also EC numbers. The action extract-annotations then allows the creation of a frequency feature table for one of these annotations within all MAGs. Currently, EC class annotations can't be selected using the flag --p-annotation and I would greatly appreciate it if they could be selected as one of the annotation options.

misialq:
Closes #276.


VinzentRisch (approved):
Hey, this looks good!
I ran it with some test MAGs and it worked fine.

PR #291, opened 2025-07-03
ENH: add new action to filter reads by metadata and remove empty samples

### LINKED ISSUE #290: ENH: filtering action for reads
- Filtering action for reads similar to what filter_mags does.
- Filter by metadata.
- Filter out empty samples.

VinzentRisch:
closes #290 

New action called `filter_reads` that can filter reads by metadata and filter out empty samples.


VinzentRisch:
Hey @misialq 
Here is the filter reads action. It can filter by metadata like the mags filter action and also remove empty files. 
I check if the files are empty by checking the first line because the size of empty fastq files was is not consistent because the length of the filename before zipping influences the size.

PR #293, opened 2025-07-04
ENH: add flag remove_empty to `filter-derep-mags` and `filter-mags`

### LINKED ISSUE #292: ENH: add flag remove-empty to filter mags
There should be a flag added to the `filter-mags` action that removes empty samples similar like it is done in the `filter-contigs` action.

VinzentRisch:
closes #292

### DIFF
--- a/q2_annotate/plugin_setup.py
+++ b/q2_annotate/plugin_setup.py
@@ -1203,14 +1203,37 @@ plugin.methods.register_function(
                 "24",
                 "25",
             ]
-        )
+        ),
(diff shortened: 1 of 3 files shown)


VinzentRisch:
Hey @misialq 
What do you think about a check if any MAGs remain after filtering? There is one of these checks in the filter-contigs action but not in the filter-samples action. And If all Mags or samples are filtered out the error is just a validation error without any information. Or is there a reason why this should not be done?

PR #318, opened 2026-01-08
ENH: add action to collapse contig abundance per annotation

### LINKED ISSUE #305: ENH: add an action to extract per-contig taxonomy
I would like a new action which can take the Kraken 2 outputs/reports and extract a per-contig taxonomy so that I can use it downstream , e.g. while trying to identify which AMR genes originated from which contig/taxon.

**Acceptance criteria**
- action name: `extract-per-contig-taxonomy`
- inputs: `SampleData[Kraken2Reports % Properties("contigs")]` + `SampleData[Kraken2Outputs % Properties("contigs")]`
- outputs: `FeatureData[Taxonomy]` where IDs correspond to contig IDs

**Additional context**
Likely we can use the existing `kraken2-to-features` action to get the taxonomies out of the report. This then needs to be combined with the contig IDs from the output files.

misialq:
Closes #305. 

Depends on https://github.com/qiime2/q2-types/pull/380.


VinzentRisch (changes requested):
Hey @misialq 
Great work! The visualisation looks nice :) I ran all actions with the data from the mock tutorial and It all runs without problems. i also checked the outputted tables and visualisation and could not find any errors.

PR #319, opened 2026-01-08
ENH: make the BUSCO visualization more responsive

### LINKED ISSUE #304: ENH: improve the BUSCO visualization
Perhaps it is not a _problem_ but the way this visualization is rendered is not optimal. The histogram plots in the first section on the first tab are overflowing the Bootstrap card if the screen is too small (honestly, it happens most of the time). The "completeness vs contamination" feels very big and zooming is a bit difficult - most of the time the contmaination is not very high but the y axis is scaled to 100% so it's hard to see the differences. On the sample details tab the plots are also overflowing to the right (fixed width?)

- the visualizations should be generated using Vega directly - Altair seems to have limitations on generating responsive plots which would fit neatly into the window width
- the plots in the "Distribution of BUSCO marker gene counts" section should be "detached" from one another so that when the window scales they can be shifted to new rows as needed
- the "Completeness vs Contamination" plot should be also a bit smaller and the axes scales should be initially scaled to the data values (start at 0 but end at 10% more than the maximum value)
- on the "Sample details" tab, the plots should also scale width-wise so that they can fit entirely within the card

misialq:
Closes #304.

### DIFF
--- a/q2_annotate/busco/busco.py
+++ b/q2_annotate/busco/busco.py
@@ -287,35 +287,16 @@ def _visualize_busco(
             dirs_exist_ok=True,
         )
 
-    # Load Vega specs from JSON files
-    vega_histogram_spec = json.dumps(_load_vega_spec("histogram"))
-    vega_unbinned_spec = json.dumps(_load_vega_spec("unbinned"))
-    vega_boxplot_spec = json.dumps(_load_vega_spec("box_plot"))
-    vega_completeness_spec = json.dumps(_load_vega_spec("completeness"))
-    vega_busco_detailed_spec = json.dumps(_load_vega_spec("busco_detailed"))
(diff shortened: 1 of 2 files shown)


q2_annotate/busco/busco.py
```diff
@@ -326,25 +429,14 @@ def evaluate_busco(
...
-            results.append(busco_result)
+    # Run BUSCO evaluation on all partitions without unbinned contigs
+    for mag_partition in partitioned_mags.values():
+        (busco_result,) = _evaluate_busco(mag_partition, db, **kwargs)
+        results.append(busco_result)
 
     (collated_results,) = collate_busco_results(results)
-    (visualization,) = _visualize_busco(collated_results)
+
+    # Pass unbinned_contigs to visualization step
```
misialq: Adding metrics for unbinned contigs was moved to the visualizer itself as all we really do here is display those on the plots. It also makes the calculation a bit easier.

PR #328, opened 2026-01-28
ENH: update extract-annotations to generate per-contig feature table

### LINKED ISSUE #311: ENH: allow grouping by either samples or features with the action qiime annotate extract-annotations
I would like to use the action qiime annotate multiply-tables to get the abundance of EC annotations from a contig frequency feature table and a table generated using qiime annotate extract-annotations. As of right now, this is not possible since qiime annotate extract-annotations can only generate a sample x annotation table from annotations generated with SampleData[Contigs] or SampleData[MAGs] artifacts. What I am proposing is an additional flag that allows you to select whether you want the table created by qiime annotate extract-annotations to be grouped by sample or feature (i.e. MAG or contig). In my use case, this would allow a link of the annotation frequency to per-sample contig abundance computed by read mapping. What I will do for now as an alternative is generate the feature table I would like outside of qiime using pandas or tidyverse and then import for use with the action qiime annotate multiply-tables.

misialq:
# Description

Closes #311. 

# AI Disclosure

- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details

Codex was used during some early stage ideating. All the code was reviewed and tested before PR submission.


q2_annotate/plugin_setup.py
VinzentRisch, line 1935: The description should be updated.

q2_annotate/eggnog/annotation.py
```diff
@@ -151,29 +126,113 @@ def _filter(data: pd.DataFrame, max_evalue: float, min_score: float) -> pd.DataF
...
+        feature_maps.append(feature_map)
         annotations.append(annot_df)
+        contig_counts.append(contig_annot_counts)
 
     result = pd.concat(annotations, axis=1).fillna(0).T
     result.index.name = "id"
-    return result
+
+    merged_maps = _merge_maps(feature_maps)
+    contig_result = pd.concat(contig_counts, axis=0).fillna(0)
```
VinzentRisch, line 236: Are contig IDs always unique?

q2_annotate/eggnog/annotation.py
```diff
@@ -151,29 +126,113 @@ def _filter(data: pd.DataFrame, max_evalue: float, min_score: float) -> pd.DataF
...
+    return feature_map, annotation_counts, contig_annotation_counts
+
+
+def _merge_maps(maps: list[dict]) -> dict:
+    """Merges a list of feature maps into a single dictionary."""
+    merged = defaultdict(set)
+    for d in maps:
+        for k, v in d.items():
+            merged[k].update(v)
+    return {k: list(v) for k, v in merged.items()}
```
VinzentRisch, line 199: Here the list of contigs could be sorted to ensure reproducibility.

VinzentRisch (changes requested):
Hello

I ran the code with some test data and it all looks good. I have some minor suggestions below:

q2_annotate/eggnog/annotation.py
misialq: Yes, contig IDs are unique across samples as we label them with UUIDs.

q2_annotate/eggnog/annotation.py
misialq: I'm not sure the tradeoff is worth it - these lists can contain thousands to hundreds of thousands of contigs per annotation so to sort all of them is not trivial. Do you have a specific "reproducibility concern" in mind?

VinzentRisch: Okay that makes sense. I have nothing in particular in mind.

VinzentRisch (approved):
Looks good :)

misialq:
Thanks @VinzentRisch! 🙏

PR #329, opened 2026-02-02
FIX: fix unclassified mags bug

colinvwood:
As noticed by a forum user [here](https://forum.qiime2.org/t/error-converting-kraken-2-report-into-taxonomy-artifact-length-of-values-does-not-match-length-of-index/33934), if all lowest common ancestor classifications for all MAGs in a dataset are to the root (i.e. unclassified), then an exception is thrown.

Instead, we now return a `[None]` from `_find_lca` instead of an empty generator expression when a root classification is encountered so that we can never inadvertently create an empty `pd.DataFrame`.


misialq:
Hey @colinvwood, this all looks good to me - could you please check out the linter errors? 🙏

Issue #322, opened 2026-01-15
ENH: the binning actions should accept sorted alignment maps


misialq:
As described in https://github.com/bokulich-lab/q2-assembly/issues/140, binning actions should not need to sort the BAM/SAM files. Instead they should receive a pre-sorted artifacts.

**Acceptance criteria:**
- remove sorting step from the metabat2 binning action
- change the input type for the alignment map to contain the "sorted" property

ebolyen:
Solved by https://github.com/bokulich-lab/q2-annotate/pull/331 and https://github.com/bokulich-lab/q2-assembly/pull/146 I believe.

PR #331, opened 2026-02-12
MAINT: remove sorting from the binning action and require sorted alignment maps

fethalen:
This PR addresses issue #322 by removing sorting from the binning with MetaBAT action, and instead requiring the alignment maps to be sorted beforehand. Sorting is now performed in a prior step (q2-assembly's `map-reads` or `sort-alignment-maps`) as a way to avoid redundant code in anticipation of adding new binners.


misialq (approved):
LGTM, thanks @fethalen! Perhaps we should wait with merging until your other PR is ready - otherwise it'd be difficult to run this 😅

fethalen:
There was an issue with one of the tests that have now been fixed!

Issue #315, opened 2025-11-05
BUG: Running `fetch-eggnog-db` results in ERROR 404: Not Found.


fethalen:
**Describe the bug**
`fetch-eggnog-db` fails to download the EggNOG database due to a broken link. This is actually not an issue in q2-annotate but with [the download_eggnog_data.py script in EggNOG](https://github.com/eggnogdb/eggnog-mapper/blob/master/download_eggnog_data.py). The issue has already been described [here](https://github.com/eggnogdb/eggnog-mapper/issues/571) and a PR fixing the broken URL can be found [here](https://github.com/eggnogdb/eggnog-mapper/pull/574). The PR has yet to be integrated into the main branch of EggNOG itself. The issue has also been mentioned in the following forum post: [https://forum.qiime2.org/t/installation-errors-in-diamond-database-and-eggnog-database/33710/16?u=nicholas_bokulich](https://forum.qiime2.org/t/installation-errors-in-diamond-database-and-eggnog-database/33710/16?u=nicholas_bokulich).

The issue will remain until the URL is updated in EggNOG. Until then, it is suggested to manually download the database and then import it as an artifact. **I have described how this is done in the last section of this issue labeled "Temporary workaround".**

**To Reproduce**

Executing the command

```bash
qiime annotate fetch-eggnog-db \
--o-db eggnog_db.qza \
--verbose
```

Produces the following error:

```text
/Users/fsandberg/miniconda3/envs/q2-annotate-2025-08-27/lib/python3.10/site-packages/q2_assembly/_utils.py:14: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
Running external command line application(s). This may print messages to stdout and/or stderr.
The command(s) being run are below. These commands cannot be manually re-run as they will depend on temporary files that no longer exist.

Command: download_eggnog_data.py -y -D --data_dir /var/folders/xn/c9590_qd0_x2pnp5ndd0fn0w0000gp/T/qiime2/fsandberg/processes/54715-1762331303.288316@fsandberg/tmp/q2-OutPath-f8i10609

[...]

PR #338, opened 2026-04-02
FIX: remove dependency on script when fetching the EggNOG database

fethalen:
This PR addresses issue #315 by removing reliance on the broken Python script. Instead, the files `eggnog.db`, `eggnog.taxa.db`, and `eggnog.taxa.traverse.pkl` are being downloaded independently using Wget.


misialq (approved):
That looks good - thanks for the quick fix @fethalen! I tested it out locally and I got my DB 💃

PR #340, opened 2026-04-30
ENH: add filter-reads-kraken2 action

misialq:
# Description

This PR introduces a new action `filter-reads-kraken2` that allows filtering a set of reads using their corresponding Kraken2 classifications.
- Taxonomy can be provided either as a literal (e.g., "Bacteria") or as an ID integer (e.g., "2").
- It is possible to filter only the indicated taxonomy or also include all of its descendants (default).
- It is possible to toggle what is being retained and discarded (by default the reads of indicated taxonomies will be retained).

# AI Disclosure

- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details

Codex was used during early stage implmentation. All the code was reviewed and tested before PR submission.


VinzentRisch:
I will review this on monday, if that is okay. I want to finish something today.

q2_annotate/kraken2/filter_reads.py
```diff
@@ -0,0 +1,547 @@
...
+    Args:
+        filepath (Path): The path to the output FASTQ file.
+        gzip_output (bool): If True, the output will be GZIP compressed.
+
+    Returns:
+        file object: A file object opened for writing in text mode.
+    """
+    if gzip_output:
+        return gzip.open(filepath, mode="wt")
+    return open(filepath, mode="w")
```
VinzentRisch: These could be one helper with the parameters for gzip and read/write.
In the code underneath you have to use _is_gzip_file anyway once outside the helpers so doing it again in _open_fastq_for_read is redundant.

q2_annotate/kraken2/filter_reads.py
```diff
@@ -0,0 +1,547 @@
...
+            "Refusing in-place overwrite to prevent data loss."
+        )
+
+    # `duplicate` may create hardlinks. Writing to a hardlinked destination
+    # would also mutate/truncate the input file.
+    if output_fp.exists() and os.path.samefile(str(input_fp), str(output_fp)):
+        raise ValueError(
+            "Input and output FASTQ paths refer to the same inode (hardlink). "
+            "Refusing in-place overwrite to prevent data loss."
+        )
```
VinzentRisch, line 126: In this implementation it looks like it is impossible to create an outputfile that is the same path as the input file becasue _filter_reads_kraken2 creates a new CasavaOneEightSingleLanePerSampleDirFmt(). 
I don't understand why this check is necessary then. The only way I can see is if you plan to use the underlying helper functions for other purposes. But for that would the "qiime way" not be to create a pipeline and use _filter_reads_kraken2 instead of the underlying helpers?

VinzentRisch (commented):
Hey Michal
I ran the code with the data from the mock tutorial and with a couple of querys. 
It all looks good but I have two comments below:

q2_annotate/kraken2/filter_reads.py
misialq: Not really - in the `write` variant of this method I cannot check whether the `filepath` is gzipped as this is an empty file. Instead, I need to know whether the input file (which is being read in the other method at the same time) is gzipped. That means that I can either pass the other file here too, or just a flag telling this method it should gzip. I don't want to make this method dependent on the second file - that does not make much sense if we can just inform it using a flag. These two methods have different behaviour and that's why I'm keeping them separate.

q2_annotate/kraken2/filter_reads.py
misialq: Technically yes, in the current implementation it should be fine. But if anyone was ever to use this method for something else (e.g., in a different action elsewhere) this condition may become relevant. During develpment this issue caused me a lot of headaches so I'd prefer to keep this check here, just in case. It's not expensive and it may prevent bugs in the future. I added a better explanation instead.

q2_annotate/kraken2/filter_reads.py
VinzentRisch: I mean it the other way around: 

```
def _open_fastq(filepath: Path, gzip: bool, mode: str):
    if gzip_output:
        return gzip.open(filepath, mode=mode + "t")
    return open(filepath, mode=mode)
```

and then use 

    gzip_input = _is_gzip_file(input_fp)
    with (
        _open_fastq(input_fp, gzip=gzip_input, mode="r") as in_fh,
        _open_fastq(output_fp, gzip=gzip_input, mode="w") as out_fh,
        
Like this _is_gzip_file(input_fp) has to be checked only once and you only need one helper.

q2_annotate/kraken2/filter_reads.py
VinzentRisch: Yes this makes sense.

q2_annotate/kraken2/filter_reads.py
misialq: Oh, of course 😅 That makes sense - I updated the code, thanks!

VinzentRisch (approved):
Yes looks good!

PR #343, opened 2026-05-21
ENH: introduce a new SemiBin2-based binning action

### LINKED ISSUE #298: ENH: add an action for contig binning using SemiBin2
Not exactly. Having a single binner available in the MOSHPIT distribution is becoming a bit limiting though.

I would like a new action `bin-contigs-semibin2` which will work similarly to the `bin-contigs-metabat` one. 

Acceptance criteria:
- the action accepts contigs and alignment maps as required inputs
- the action can perform both, sinle-sample and multi-sample, bininng - configurable by the user (single-sample is default)
- the action uses SemiBin2
- the action outputs the formed bins before and after reclustering and the contig map, similarly like it's done in the existing binning action
- the action exposes parameters like `threads`, `min-len`, `batch-size`, `epochs`, `random-seed` and `sequencing-type`
- the action always runs SemiBin2 in the verbose mode
- the conda recipe meta.yaml is updated accordingly to include the semibin package
- all the functionality is unit tested

**Additional context**
Usage examples: https://github.com/BigDataBiology/SemiBin
Documentation: https://semibin.readthedocs.io/en/latest/

fethalen:
# Description

This PR addresses issue #298 by introducing a new SemiBin2-based MAG binning action to q2-annotate.

- The action accepts contigs and alignment maps as required inputs
- At the moment of writing, the action can perform only single-sample binning
- The action outputs the formed bins before and after reclustering and the contig map, similar to the existing binning action
- The action exposes parameters like threads, min-len, batch-size, epochs, random-seed and sequencing-type
- The action always runs SemiBin2 in the verbose mode
- The conda recipe meta.yaml has been updated to include the semibin package
- All the functionality is unit tested 😎

# AI Disclosure

- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details

ChatGPT's GPT‑5.5 was used to assist with tasks such as code refinement and prototyping tests.


q2_annotate/semibin2/semibin2.py
```diff
@@ -0,0 +1,127 @@
...
+        raise ValueError(
+            "No MAGs were formed during binning, please check your inputs."
+        )
+
+    contig_map = _generate_contig_map(bins)
+
+    return bins, contig_map
+
+
+def bin_contigs_semibin2(
```
misialq, line 86: Normally, we set the default values based on what the underlying tool uses as defaults. Could you please update this signature?

q2_annotate/semibin2/semibin2.py
```diff
@@ -0,0 +1,127 @@
...
+    ml_threshold: int | None = None,
+    num_threads: int | None = None,
+    seed: int | None = None,
+    debug: bool | None = None,
+    verbose: bool | None = None,
+) -> (MultiFASTADirectoryFormat, dict):
+    kwargs = {
+        k: v
+        for k, v in locals().items()
+        if k not in ["contigs", "alignment_maps", "mode", "training_type", "verbose"]
```
misialq: Including `verbose` here means you are actually not passing it anywhere afterwards, right?

q2_annotate/semibin2/semibin2.py
misialq, line 86: Those parameters will not work as they do not reflect the names that SemiBin2 is using:
- `seed` (instead of `random_seed`)
- `num_threads` (instead of `threads`)

conda-recipe/meta.yaml
```diff
@@ -22,6 +22,7 @@ requirements:
   - kaiju
   - kraken2
   - metabat2
+  - semibin
```
misialq: This should be using a pin so that the exact version is picked up from the seed environment (like busco for example).

misialq (changes requested):
Hey @fethalen, here is my first superficial review. I'm gonna test it out with some data and come back with more feedback later. 

Also, can you update the PR description to match out new template? 🙏

misialq (commented):
Hey @fethalen, I could successfully generate some bins for a couple of samples. I also ran BUSCO through them and compared to the bins obtained from metabat2 - the results look similar. So this PR will be good to go once you address the other comments (e.g., I was not able to set the number of cpus used by semibin - it always defaults to all available).

fethalen:
> Hey @fethalen, I could successfully generate some bins for a couple of samples. I also ran BUSCO through them and compared to the bins obtained from metabat2 - the results look similar. So this PR will be good to go once you address the other comments (e.g., I was not able to set the number of cpus used by semibin - it always defaults to all available).

HI @misialq and thank you for your comments. I've tried to address them all as best as I can, please let me know if you have any questions.

misialq (approved):
Hey @fethalen, that's great, thanks! 🚀

PR #346, opened 2026-06-03
FIX: paginate results when fetching Kraken 2 DBs

misialq:
# Description

This PR fixes a bug by which the S3 results returned within the `fetch-kraken-db` action would be limited to the first 1000 keys. This would result in the action failing as the Kraken databases recently stopped being included in the first 1000 results. Instead, we should paginate the results, fetch all and search for the required collection through those.

To test simply run the action with any of the collection keys - the issue applied to all of those.

# AI Disclosure

- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details

Gemini was used while updating tests and adding new ones. All the code was reviewed before PR submission.


fethalen (approved):
Hi @misialq, this is looking good to me and fixes a seemingly important issue. I've tested the `build-kraken-db` action using the pluspf collection and it works fine on my end.

PR #349, opened 2026-06-08
ENH: add an action to transfer functional annotations

### LINKED ISSUE #341: ENH: add an action to transfer functional annotations
In the current MAG recovery flow the set of contigs between `SampleData[Contigs]` and `FeatureData[MAGs]` gets progressively reduced while preserving contig IDs. If we annotated the `SampleData[Contigs]` using `map-eggnog` it should technically be possible to "transfer" the annotations obtained for this set of contigs to any downstream set of contigs. This would greatly reduce the resource usage for the users who want to annotate each of those inputs.

We should introduce a new action `transfer-eggnog-annotations` which would perform the transfer between compatible inputs:
- `SampleData[Contigs]` -> `SampleData[MAGs]`
- `SampleData[MAGs]` -> `FeatureData[MAG]` (this is actually a simple filter action on the superset of MAGs from the input)
- `SampleData[Contigs]` -> `FeatureData[MAG]`

The action should accept two inputs:
- the eggnog annotations of the "source"
- the sequences of the "destination" (one of the semantic types above)
 
There should be one output containing eggnog annotations matching the "destination" sequences.

ZuzanaSebb:
# Description
Add  transfer_eggnog_annotations  action.

## What it does

- MAG-level annotations + `FeatureData[MAG]` → copy files for matching MAG IDs (after dereplication).
- Contig-level annotations + `FeatureMap[MAGtoContigs]` → aggregate into per-MAG files by mapping each gene to its contig.

# AI Disclosure
- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details
Claude (Sonnet 4.6) was used to draft the initial implementation; I refined the methods and the final action.
Claude (Opus 4.8) was used to draft the initial implementation of the unit tests.


q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +185,126 @@ def extract_annotations(
     result = pd.concat(annotations, axis=1).fillna(0).T
     result.index.name = "id"
     return result
+
+
+def _get_mag_ids_from_feature_data(mags: MAGSequencesDirFmt) -> set:
```
misialq: Do we really need this method? It literally does a single thing.

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +185,126 @@ def extract_annotations(
...
+    """Extract MAG UUIDs from a FeatureData[MAG] artifact."""
+    return set(mags.feature_dict().keys())
+
+
+def _copy_annotation_files(
+    source_annotations: OrthologAnnotationDirFmt,
+    mag_ids: set,
+    result: OrthologAnnotationDirFmt,
```
misialq: You should just make this method return `results` as nothing really happens to them before they are being passed to this method. This also makes it a bit more explicit what this method actually does/returns.

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +185,126 @@ def extract_annotations(
...
+):
+    """Copy annotation files from source to result for the given MAG IDs."""
+    annotation_dict = source_annotations.annotation_dict()
+
+    matched_ids = mag_ids & set(annotation_dict.keys())
+    if not matched_ids:
+        raise ValueError("No annotation files matched the destination MAG IDs.")
```
misialq: I think this check should happen already before we call this method - it will remove the need to pass the `mag_ids` and provide a simpler interface. Actually, you could even move it to a separate method as you have some additional check below - that way the validation can be taken care of one testable method and the copying by another one, making both of them responsible for different parts of the pipeline.

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +185,126 @@ def extract_annotations(
...
+        f"row(s) into {matched['mag_uuid'].nunique()} MAG(s); "
+        f"{unmatched} row(s) skipped."
+    )
+
+    return result
+
+
+def transfer_eggnog_annotations(
+    ortholog_annotations: OrthologAnnotationDirFmt,
+    destination: Union[MAGSequencesDirFmt, MAGtoContigsDirFmt],
```
misialq: This is confusing. The destination should be represented by the same kind of semantic type, either SampleData[MAGs] or FeatureData[MAG]. Now, you are mixing in the contig map. I think this should become an additional input required when SampleData[MAGs] were provided as source of the annotations or if FeatureData[MAG] was provided as the destination (whichever of those makes more sense for your pipeline).

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +185,126 @@ def extract_annotations(
...
+    }
+
+    # Read all annotation files into a DataFrame
+
+    frames = []
+    for _id, fp in ortholog_annotations.annotation_dict().items():
+        df = pd.read_csv(fp, sep="\t", skiprows=4)
+        # drop trailing comment only if present
+        first_col = df.columns[0]
+        df = df[~df[first_col].astype(str).str.startswith("##")]
```
misialq: I'm wondering whether you could achieve the same by simply reading in every file as OrthologFileFmt and `view`ing as a df?

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +185,126 @@ def extract_annotations(
...
+        frames.append(df)
+
+    all_annotations = pd.concat(frames, ignore_index=True)
```
misialq: We need to evaluate carefully whether this will work well when one has hundreds of samples with thousands of annotations. I'm a bit worried that the memory will blow up 😅

misialq (changes requested):
Hey @ZuzanaSebb, here's my first round of comments - once you update these I'll take your code for a spin :D

Also, just a reminder, in all the plugin repos in our organization we are trying to follow the PR naming scheme same as commit naming (use a prefix and a semi-verbose description, typically using a verb in imperative form).

q2_annotate/eggnog/annotation.py
ZuzanaSebb: we could use `_annotations_to_dataframe` fuction but the result would need to be adjusted anyway

misialq:
Hey @ZuzanaSebb, it seems like there are some unresolved conflicts - could you please fix those before I review? 🙏

q2_annotate/plugin_setup.py
```diff
@@ -1817,6 +1817,42 @@
     citations=[],
 )
 
+plugin.methods.register_function(
```
misialq: This should become a pipeline so that we can detect which inputs were provided (see my other comment in the tranfser code).

q2_annotate/eggnog/annotation.py
misialq: So did you try to evaluate this a little bit?

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +188,178 @@ def extract_annotations(
...
+    return result
+
+
+def transfer_eggnog_annotations(
+    ortholog_annotations: OrthologAnnotationDirFmt,
+    destination: Union[MAGSequencesDirFmt, MultiMAGSequencesDirFmt],
+    contig_map: MAGtoContigsDirFmt = None,
+) -> OrthologAnnotationDirFmt:
+    """Transfer or aggregate eggNOG annotations based on source and destination."""
+    if _annotations_from_contigs(ortholog_annotations):
```
misialq: When this is a `pipeline` you won't need this helper - you can directly check what the semantic type of the input artifact is (see how it's done in the `classify-kraken2` pipeline).

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +188,178 @@ def extract_annotations(
...
+        f"row(s) into {matched['mag_uuid'].nunique()} MAG(s); "
+        f"{unmatched} row(s) skipped.",
+        UserWarning,
+    )
+
+    return result
+
+
+def transfer_eggnog_annotations(
+    ortholog_annotations: OrthologAnnotationDirFmt,
```
misialq: Let's call this `source_annotations`

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +188,178 @@ def extract_annotations(
...
+    destination: Union[MAGSequencesDirFmt, MultiMAGSequencesDirFmt],
```
misialq: `destination_sequences`

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +188,178 @@ def extract_annotations(
...
+    contig_map: MAGtoContigsDirFmt = None,
```
misialq: `source_contig_map`

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +188,178 @@ def extract_annotations(
...
+
+def _annotate_mags_from_contigs(
+    ortholog_annotations: OrthologAnnotationDirFmt,
+    destination: Union[MAGSequencesDirFmt, MultiMAGSequencesDirFmt],
+    contig_map: MAGtoContigsDirFmt = None,
+) -> OrthologAnnotationDirFmt:
+    """Aggregate contig-level eggNOG annotations -> MAG-level annotations."""
+
+    if contig_map is not None:
+        flat_contig_map = contig_map.file.view(dict)
```
misialq: If you need to work with a `dict` you should be able to set it as a type annotation on the main pipeline function - the transformer will then give you the dict directly and you won't need to convert here.

q2_annotate/eggnog/annotation.py
```diff
@@ -177,3 +188,178 @@ def extract_annotations(
...
+        try:
+            is_uuid4 = str(UUID(value, version=4)) == value
+        except ValueError:
+            is_uuid4 = False
+        if not is_uuid4:
+            return True
+    return False
+
+
+def _mag_ids(
```
misialq: `_get_mag_ids`

q2_annotate/eggnog/annotation.py
```diff
@@ -231,8 +244,181 @@ def extract_annotations(
 
     result = pd.concat(annotations, axis=1).fillna(0).T
     result.index.name = "id"
-
     merged_maps = _merge_maps(feature_maps)
```
misialq, line 238: Please add a test for both destination types.

q2_annotate/eggnog/annotation.py
```diff
@@ -231,8 +244,180 @@ def extract_annotations(
...
     contig_result = pd.concat(contig_counts, axis=0).fillna(0)
-
     return result, dict(merged_maps), contig_result
+
+
+def _validate_mag_ids(mag_ids: set, annotation_dict: dict) -> set:
```
misialq: Including type names in the variable is considered a Python anti-pattern and should be avoided - that's what the type annotations are for. Please review your code and rename the variables accordingly (e.g., here let's just call it `annotations`)

q2_annotate/eggnog/annotation.py
```diff
@@ -231,8 +244,180 @@ def extract_annotations(
...
+
+    matched = all_annotations.dropna(subset=["mag_uuid"])
+
+    if matched.empty:
+        raise ValueError("No annotation rows could be matched to any MAG.")
+
+    unmatched = len(all_annotations) - len(matched)
+    if unmatched > 0:
+        total = len(all_annotations)
+        pct = unmatched / total * 100
```
misialq: pct?

q2_annotate/eggnog/annotation.py
```diff
@@ -231,8 +244,180 @@ def extract_annotations(
...
+        mags = destination.feature_dict()
+
+    contig_map = {}
+    for mag_id, mag_fp in mags.items():
+        seqs = skbio.read(str(mag_fp), format="fasta", verify=False)
+        contig_map[mag_id] = [x.metadata["id"] for x in seqs]
+    return contig_map
+
+
+def _annotate_mags_from_contigs(
```
misialq: This method is responsible for a lot of functionality - please break it down into smaller, testable units and add unit tests for all of them.

q2_annotate/plugin_setup.py
```diff
@@ -1322,6 +1323,43 @@
...
+    function=q2_annotate.eggnog.transfer_eggnog_annotations,
+    inputs={
+        "ortholog_annotations": GenomeData[NOG],
+        "destination": FeatureData[MAG] | SampleData[MAGs],
+        "contig_map": FeatureMap[MAGtoContigs],
+    },
```
misialq, line 1360: You should use the `TypeMap` here to control which pairs of inputs are allowed - you can see some examples in this repo (e.g., some Kraken2 actions).

q2_annotate/eggnog/annotation.py
misialq, line 1: I think all of this functionality should go into its own submodule, perhaps `transfer.py`?

misialq (changes requested):
Hey @ZuzanaSebb, here's some more work for you! 😄

q2_annotate/eggnog/annotation.py
ZuzanaSebb: I tested on a larger data set, containing 36 samples with ~ 1.5M annotation rows and it completed without any memory issues.

q2_annotate/eggnog/annotation.py
ZuzanaSebb: I can't get rid of the `_annotations_from_contigs` helper because `GenomeData[NOG]` doesn't have Properties to distinguish contigs vs MAG-level annotations.

misialq: Oh no, that's unfortunate 🤦 Can you please open an issue in this repo to add those properties? That could help us in situations like this one... I think we should still be able to add that for you to refactor your code before the release - I really think this check should be unnecessary.

ZuzanaSebb: Sure, I think after that the TypeMap for different inputs makes more sense as well.

q2_annotate/eggnog/annotation.py
```diff
@@ -216,7 +218,9 @@ def extract_annotations(
         annot_df = pd.read_csv(
             fp, sep="\t", skiprows=4, index_col=0
         )  # skip the first 4 rows as they contain comments
-        annot_df = annot_df.iloc[:-3, :]  # remove the last 3 comment rows
+        # strip trailing comment rows (footer) only if present
+        if annot_df.index[-3:].astype(str).str.startswith("##").all():
```
misialq, line 222: Why did this need to change? Is it because when we transfer those lines would be missing?

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,266 @@
...
+    destination_sequences,
+    source_contig_map=None,
+):
+    """Transfer or aggregate eggNOG annotations based on source and destination."""
+    source_is_contig_typed = _is_source_contig_typed(source_annotations)
+
+    if not source_is_contig_typed and source_contig_map is not None:
+        raise ValueError(
+            "`source_contig_map` is only valid for contig-level source annotations."
+        )
```
misialq: This should be possible to achieve with the TypeMatch, I think - if contigs are provided then the contig_map has to be None.

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,266 @@
...
+
+    source_fmt = source_annotations.view(OrthologAnnotationDirFmt)
```
misialq: Let's not call these `*_fmt` - this is a bit confusing. Maybe you can just simplify to "source" and "destination"?

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,266 @@
...
+def _transfer_eggnog_annotations(
+    source_annotations: OrthologAnnotationDirFmt,
+    destination_sequences: Union[MAGSequencesDirFmt, MultiMAGSequencesDirFmt],
+    source_contig_map: dict = None,
+    is_contig_typed: bool = False,
+) -> OrthologAnnotationDirFmt:
+    if is_contig_typed:
+        return _annotate_mags_from_contigs(
+            source_annotations, destination_sequences, source_contig_map
+        )
```
misialq: This is not really the QIIME way. When you have the pipeline you could coudl just make two separate (hidden) actions - one that transfers from contigs and another one that transfers from mags. This way you won't need to pass around those weird variables like "is_contig_typed".

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,266 @@
...
+            fh.write("## Source: contig-level annotations\n")
+            fh.write(f"## MAG: {mag_uuid} | contigs: {n_contigs} | rows: {n_rows}\n")
+            fh.write("##\n")
+            fh.write(col_header)
+            group[data_cols].to_csv(fh, sep="\t", index=False, header=False)
+
+    return result
+
+
+def _annotate_mags_from_contigs(
```
misialq: This can become a helper `action`.

misialq (changes requested):
Hey @ZuzanaSebb, you should turn some of the helper methods in your code into actual `actions` and call those from within the pipeline - see below.

q2_annotate/plugin_setup.py
```diff
@@ -1336,6 +1337,117 @@
...
+    name="Aggregate contig-level eggNOG annotations onto MAGs.",
+    description=(
+        "Aggregates contig-level eggNOG ortholog annotations into per-MAG "
+        "files using a contig-to-MAG mapping."
+    ),
+    citations=[],
+)
+
+plugin.methods.register_function(
+    function=q2_annotate.eggnog._copy_mag_annotations,
```
misialq: Why don't you just call it "_annotate_mags_from_mags" or something like that to mirror the other hidden method? Or maybe we should name both to something like:
- `_transfer_annotations_from_contigs`
- `_transfer_annotations_from_mags`

to avoid the double-mag text in the name.

q2_annotate/plugin_setup.py
```diff
@@ -1336,6 +1337,117 @@
...
+        "source_annotations": GenomeData[NOG % Properties("mags")],
+        "destination_sequences": I_transfer_destination,
+    },
+    parameters={},
+    outputs=[("transferred_annotations", GenomeData[NOG % Properties("mags")])],
+    input_descriptions={
+        "source_annotations": "MAG-level ortholog annotations to transfer.",
+        "destination_sequences": (
+            "MAGs to transfer annotations onto. Accepts either "
+            "SampleData[MAGs] or FeatureData[MAG] (e.g. dereplicated MAGs)."
```
misialq: Please remove this text from the description - this will be added by qiime automatically.

q2_annotate/plugin_setup.py
```diff
@@ -1336,6 +1337,117 @@
...
+        ),
+    },
+    parameter_descriptions={},
+    output_descriptions={
+        "transferred_annotations": "Transferred per-MAG annotations.",
+    },
+    name="Transfer MAG-level eggNOG annotations onto matching MAGs.",
```
misialq: Sometimes you refer to them as "ortholog annotations" and some other times "eggnog annotations". Let's stick with one (I'd go for the eggnog version).

q2_annotate/plugin_setup.py
```diff
@@ -1336,6 +1337,117 @@
...
+        "destination_sequences": I_transfer_destination,
+        "source_contig_map": FeatureMap[MAGtoContigs],
+    },
+    parameters={},
+    outputs=[("transferred_annotations", GenomeData[NOG % Properties("mags")])],
+    input_descriptions={
+        "source_annotations": "Contig-level ortholog annotations to aggregate.",
+        "destination_sequences": (
+            "MAGs to aggregate annotations onto. Accepts either "
+            "SampleData[MAGs] or FeatureData[MAG] (e.g. dereplicated MAGs)."
```
misialq: Same as below.

q2_annotate/plugin_setup.py
```diff
@@ -1336,6 +1337,117 @@
...
+    function=q2_annotate.eggnog.transfer_eggnog_annotations,
+    inputs={
+        "source_annotations": I_source_annotations,
+        "destination_sequences": I_destination_sequences,
+        "source_contig_map": FeatureMap[MAGtoContigs],
+    },
+    parameters={},
+    outputs=[("transferred_annotations", O_transferred_annotations)],
+    input_descriptions={
+        "source_annotations": "Ortholog annotations to transfer or aggregate.",
```
misialq: I'm wondering whether using "aggregation" here is a bit confusing. Even though I see how you mean this to be an aggregation operation, maybe we should stick with calling all of those operations simply "transfer" to avoid any confusion. That's what it ultimately is, just taking a set of annotations and _transferring_ them into a different set of contigs.

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,231 @@
...
+    return _copy_annotation_files({k: annotations[k] for k in matched_ids})
+
+
+def transfer_eggnog_annotations(
+    ctx,
+    source_annotations,
+    destination_sequences,
+    source_contig_map=None,
+):
+    """Transfer or aggregate eggNOG annotations based on source and destination."""
```
misialq: See my comment in plugin_setupy.py.

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,231 @@
...
+    if source_annotations.type <= GenomeData[NOG % Properties("contigs")]:
+        transfer_action = ctx.get_action("annotate", "_annotate_mags_from_contigs")
+        (transferred_annotations,) = transfer_action(
+            source_annotations, destination_sequences, source_contig_map
+        )
+    else:
+        if source_contig_map is not None:
+            raise ValueError(
```
misialq: Do we need to error here? Perhaps we should just raise a warning and say that these will be ignored not to break the flow.

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,231 @@
...
+                "`source_contig_map` is only valid for contig-level source "
+                "annotations."
+            )
+        transfer_action = ctx.get_action("annotate", "_copy_mag_annotations")
+        (transferred_annotations,) = transfer_action(
+            source_annotations, destination_sequences
+        )
```
misialq: Instead of calling the `transfer_function` in each branch, just set the `source_contig_map` to None in the second branch and call the method after the if statement.

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,231 @@
...
+    total = len(all_annotations)
+    return matched, total
+
+
+def _warn_unmatched_annotation_rows(unmatched: int, total: int) -> None:
+    if unmatched > 0:
+        pct = unmatched / total * 100
+        warnings.warn(
+            f"{unmatched} of {total} annotation row(s) ({pct:.1f}%) were on "
+            "contigs not present in the contig map (e.g. unbinned contigs) "
```
misialq: I am confused by this message - why do you say here "e.g. unbinned contigs"? These contigs are not part of the contig map nor are they part of the downstream artifacts so it is not possible that unmatched annotations would arise due to unbinned contigs being present. Or did I miss something? When does this situation actually happen? I mean, it's good to have this warning jsut in case but the message needs to make sense to what is actually possible (and that's what I'm trying to understand).

q2_annotate/eggnog/tests/test_transfer.py
misialq, line 1: I just skimmed through your tests and have one, high-level comment. Ideally, we never generate any test data on the fly. All the test data should be placed in FASTA files committed in this repo so that anyone can look at them and understand what's going on. Needing to go through your 4 (!) functions that just build the destination sequences increases the test complexity unnecessarily and makes the tests less reliable. Please come up with a minimal dataset and commit the respective files. I will review this code only once you have added those.

misialq (changes requested):
Another round!

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,231 @@
...
+        if source_contig_map is not None:
+            warnings.warn(
+                "`source_contig_map` is only valid for contig-level source "
+                "annotations and will be ignored.",
+                UserWarning,
+            )
+        transfer_action = ctx.get_action("annotate", "_transfer_annotations_from_mags")
+        args = (source_annotations, destination_sequences)
+
+    (transferred_annotations,) = transfer_action(*args)
```
misialq, line 241: This could be even simpler:
```suggestion
    if source_annotations.type <= GenomeData[NOG % Properties("contigs")]:
        transfer_action = ctx.get_action(
            "annotate", "_transfer_annotations_from_contigs"
        )
    else:
        if source_contig_map is not None:
            warnings.warn(
                "`source_contig_map` is only valid for contig-level source "
                "annotations and will be ignored.",
                UserWarning,
            )
        transfer_action = ctx.get_action("annotate", "_transfer_annotations_from_mags")
        source_contig_map = None

    (transferred_annotations,) = transfer_action(source_annotations, destination_sequences, source_contig_map)
```

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+        self.expected_contig_map = {
+            MAG1: ["mtjebimcR24S9DZ62TY6Fh", "NNqHkme8fLmessev7CnUMU"],
+            MAG2: ["ipio8kS3aBF5G9Lw6XsSxj"],
+        }
+
+    def _query_ids(self, result, mag_uuid):
+        df = pd.read_csv(result.annotation_dict()[mag_uuid], sep="\t", skiprows=4)
+        return df[df.columns[0]].tolist()
+
+    def test_build_contig_map(self):
```
misialq: This should be split into separate tests - it makes it easier to notice the issues if only one of the two fails (instead of digging through the tests to figure out what happened).

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+        contig_map_from_derp_mags = _build_contig_map(self.destination_sequences)
```
misialq:
```suggestion
        contig_map_from_derep_mags = _build_contig_map(self.destination_sequences)
```

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+        contig_map_from_mags_per_sample = _build_contig_map(
+            self.destination_sequences_per_sample
+        )
+
+        self.assertEqual(
+            contig_map_from_derp_mags,
```
misialq:
```suggestion
            contig_map_from_derep_mags,
```

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+    def test_get_mag_ids(self):
+        mag_ids_from_derp_mags = _get_mag_ids(self.destination_sequences)
+        mag_ids_from_mags_per_sample = _get_mag_ids(
+            self.destination_sequences_per_sample
+        )
+        expected_mag_ids = {MAG1, MAG2}
+        self.assertEqual(mag_ids_from_derp_mags, expected_mag_ids)
+        self.assertEqual(mag_ids_from_mags_per_sample, expected_mag_ids)
+
+    def test_resolve_contig_map(self):
```
misialq: Same here: split into two tests.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+    def test_reverse_contig_map(self):
+        source_contig_map = {
+            MAG1: ["mtjebimcR24S9DZ62TY6Fh", "NNqHkme8fLmessev7CnUMU"],
+            MAG2: ["ipio8kS3aBF5G9Lw6XsSxj"],
+        }
+        contig_to_mag, n_contigs_by_mag = _reverse_contig_map(source_contig_map)
+        self.assertEqual(contig_to_mag["mtjebimcR24S9DZ62TY6Fh"], MAG1)
+        self.assertEqual(contig_to_mag["ipio8kS3aBF5G9Lw6XsSxj"], MAG2)
+        self.assertEqual(n_contigs_by_mag[MAG1], 2)
+        self.assertEqual(n_contigs_by_mag[MAG2], 1)
```
misialq: Yes, Claude and Co. love doing that. Instead of making all these assertions please just construct the full dictionary you're expecting to get and `assertDictEquals` once.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+    def test_map_rows_to_mag_ids(self):
+        obs = _map_rows_to_mag_ids(self.test_annotations_df, self.test_contig_to_mag)
+        self.assertEqual(obs.loc[0, "mag_uuid"], MAG1)
+        self.assertEqual(obs.loc[1, "mag_uuid"], MAG2)
+        self.assertTrue(pd.isna(obs.loc[2, "mag_uuid"]))
+
+    def test_require_matched_annotation_rows(self):
+        tagged = self.test_annotations_df.assign(mag_uuid=[MAG1, MAG2, None])
+        matched, total = _require_matched_annotation_rows(tagged)
+        self.assertEqual(len(matched), 2)
```
misialq: Why don't you assert the actual content instead?

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+    def test_require_matched_annotation_rows_raises_on_empty(self):
+        tagged = self.test_annotations_df.assign(mag_uuid=None)
+        with self.assertRaisesRegex(ValueError, "No annotation rows could be"):
+            _require_matched_annotation_rows(tagged)
+
+    def test_warn_unmatched_annotation_rows_warns_when_nonzero(self):
+        tagged = self.test_annotations_df.assign(mag_uuid=[MAG1, MAG2, None])
+        matched, total = _require_matched_annotation_rows(tagged)
+        unmatched = total - len(matched)
+        with self.assertWarns(UserWarning):
```
misialq: Why don't you assert the message itself?

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+            }
+        )
+        result = _write_grouped_annotations(
+            matched,
+            {MAG1: 2, MAG2: 1},
+        )
+        lines = Path(result.annotation_dict()[MAG1]).read_text().splitlines()
+        self.assertIn("## Source: contig-level annotations", lines[1])
+        self.assertIn("#query\tseed_ortholog", lines[4])
+        self.assertIn("mtjebimcR24S9DZ62TY6Fh_0\tortholog1", lines[5])
```
misialq: And what happened to the other MAG?

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def test_aggregate_contigs_into_mags(self):
```
misialq: Again, please split into separate tests - if something breaks here it'll be so much easier to debug.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+class TestTransferEggnogAnnotationsPipeline(TestPluginBase):
+    package = "q2_annotate.eggnog.tests"
+
+    def setUp(self):
+        super().setUp()
+        self.transfer_eggnog_annotations = self.plugin.pipelines[
+            "transfer_eggnog_annotations"
+        ]
+
+    def test_transfer_annotations_from_mags(self):
```
misialq: And AGAIN, split into independent tests.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,399 @@
...
+            UserWarning, "only valid for contig-level source annotations"
+        ):
+            (result,) = self.transfer_eggnog_annotations(
+                source, mag_sequences_artifact, contig_map
+            )
+        self.assertTrue(result.type <= GenomeData[NOG % Properties("mags")])
+        obs = result.view(OrthologAnnotationDirFmt)
+        self.assertEqual(set(obs.annotation_dict().keys()), {MAG1, MAG2})
+
+    def test_transfer_annotations_from_contigs(self):
```
misialq: For the final time, two tests.

misialq (changes requested):
These tests still need some more work 🧰

q2_annotate/eggnog/transfer.py
misialq: Wanna give this a go?

q2_annotate/eggnog/transfer.py
```diff
@@ -0,0 +1,231 @@
...
+        raise ValueError("No annotation rows could be matched to any MAG.")
+    total = len(all_annotations)
+    return matched, total
+
+
+def _warn_unmatched_annotation_rows(unmatched: int, total: int) -> None:
+    if unmatched > 0:
+        pct = unmatched / total * 100
+        warnings.warn(
+            f"{unmatched} of {total} annotation row(s) ({pct:.1f}%) could not be "
```
misialq:
```suggestion
        frac_unmatched = unmatched / total * 100
        warnings.warn(
            f"{unmatched} of {total} annotation row(s) ({frac_unmatched:.1f}%) could not be "
```

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,439 @@
...
+        self.assertEqual(
+            resolved_map_from_mags_per_sample,
+            self.expected_contig_map,
+        )
+
+    def test_load_annotation_rows(self):
+        obs = _load_annotation_rows(self.source_annotations)
+        self.assertEqual(len(obs), 10)
+        self.assertFalse(obs[obs.columns[0]].astype(str).str.startswith("##").any())
+        self.assertIn("pXaG7nQ3mZtY8LbKdWs1Rf_1", obs[obs.columns[0]].tolist())
```
misialq: Since you know you expect to get a DataFrame you could also construct the expected DataFrame, dump it to TSV file and read it back in here for comparison. If the file you are using has a lot of annotation values you can also simplify it a bit by removing some of them.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,439 @@
...
+                "ipio8kS3aBF5G9Lw6XsSxj": MAG2,
+            },
+        )
+        self.assertDictEqual(n_contigs_by_mag, {MAG1: 2, MAG2: 1})
+
+    def test_map_rows_to_mag_ids(self):
+        obs = _map_rows_to_mag_ids(self.test_annotations_df, self.test_contig_to_mag)
+        self.assertEqual(obs.loc[0, "mag_uuid"], MAG1)
+        self.assertEqual(obs.loc[1, "mag_uuid"], MAG2)
+        self.assertTrue(pd.isna(obs.loc[2, "mag_uuid"]))
```
misialq: This is a bit brittle as it relies on indexing. A better way would be to assert the entire column is correct - you can create a series that is expected and compare against it instead.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,439 @@
...
+            f"mtjebimcR24S9DZ62TY6Fh_0\tortholog1\n"
+        )
+        expected_MAG2 = (
+            f"## Transferred using transfer_eggnog_annotations (q2-annotate)\n"
+            f"## Source: contig-level annotations\n"
+            f"## MAG: {MAG2} | contigs: 1 | rows: 1\n"
+            f"##\n"
+            f"#query\tseed_ortholog\n"
+            f"ipio8kS3aBF5G9Lw6XsSxj_0\tortholog2\n"
+        )
```
misialq: That's exactly what I was hoping you wouldn't do - the method you are testing creates these files so why don't you just put those as files in the test data and do file comparisons instead? It's also much easier to compare "by eye" just by looking at two sets of files.

q2_annotate/eggnog/tests/test_transfer.py
```diff
@@ -0,0 +1,439 @@
...
+        self.assertEqual(
+            Path(result.annotation_dict()[MAG1]).read_text(), expected_MAG1
+        )
+        self.assertEqual(
+            Path(result.annotation_dict()[MAG2]).read_text(), expected_MAG2
+        )
+
+    def _assert_transfer(self, destination_sequences, source_contig_map=None):
```
misialq: This is similar to what I pointed out above: this kind of a test should rely on a set of input files and corresponding, expected output files - you should compare both in the test. These are then also much easier to compare by eye, if need be.

q2_annotate/eggnog/tests/test_transfer.py
ZuzanaSebb: I added 2 more files inside of a new folder expected-grouped-annotations.

q2_annotate/eggnog/tests/test_transfer.py
ZuzanaSebb: done!

q2_annotate/eggnog/tests/test_transfer.py
ZuzanaSebb: I created a tsv file and kept all the columns as in the other annotations folder.

misialq (changes requested):
Hey @ZuzanaSebb, I tired it again after your latest fix but the diff still looks different. What we want is that the file after transfer looks exactly the same as the one generated by EggNOG (except of the comment lines since we are adding those "manually"). You can just run the _diff_ tool yourself to see what I mean - I only want to see the comments marked as being different. You can see now why I was insisting on using file comparisons in the tests: please change your tests so that they compare the exact file contents - this way we don't have to run those diffs every time. Also, ideally, please use some data that is "real": you can obviously trim down the files not to test thousands of lines and annotations. Thanks!

ZuzanaSebb:
Hi @misialq! Thanks for testing. To achieve the same written form I decided to import the values as type string to avoid this floating point precision problems. Also the problem was hidden in the `pd.testing.asset_frame_equal` doesn't  fail comparing same floats of different written form. This should be avoided now as all the columns are imported as strings.

misialq (approved):
Hey @ZuzanaSebb, perfect, thanks! I tried this out and made a bunch of comparisons - it all seems to be working fine!

PR #359, opened 2026-07-10
IMP: add properties on the eggnog outputs to make them more specific

### LINKED ISSUE #358: ENH: add Properties for GenomeData[NOG]
When implementing the `transfer-eggnog-annotations` pipeline in q2-annotate, we need to distinguish between contig-level and MAG-level eggNOG annotations at the semantic type level. Currently both are typed as GenomeData[NOG].

Add Properties support to the GenomeData[NOG] semantic type so that annotations derived from contigs and MAGs can be typed differently.

```
GenomeData[NOG % Properties("contigs")]  # annotations from SampleData[Contigs]
GenomeData[NOG % Properties("mags")]     # annotations from SampleData[MAGs]
```

Currently we inspect annotation filenames to determine the source type — sample names (e.g. c2_e4) indicate contig-level annotations while UUIDs indicate MAG-level annotations.

misialq:
# Description

Closes #358.

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

### DIFF
--- a/q2_annotate/eggnog/tests/test_annotation.py
+++ b/q2_annotate/eggnog/tests/test_annotation.py
@@ -54,7 +54,8 @@ class TestAnnotate(TestPluginBase):
 
     def test_eggnog_annotate_parallel(self):
         orthologs = qiime2.Artifact.import_data(
-            "SampleData[Orthologs]", self.get_data_path("good_hits/")
+            "SampleData[Orthologs % Properties('contigs')]",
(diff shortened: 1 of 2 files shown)


ZuzanaSebb:
I tested your PR changes together with the corresponding updates in q2-types. I verified that both `search-orthologs-diamond` and `map-eggnog` return the correct semantic types with Properties

